PR #2, opened 2018-02-05
Feature: Bidirectional(ish) inference

zardyh:
Move to a pseudo-bidirectional type system algorithm, similar to the
one described for Damas-Milner inference in [0]. This isn't sound at
all, yet, particularly because of fiddlyness around the subsumption
check.b

We reuse most of the infrastructure, but the core judgement is
`Γ ⊢ x <= t` instead of `Γ ⊢ x => t` (i.e. checking instead of
synthesis).

You can recover inference from checking in a unification-based
system (like ours) like this:

```
infer e = do
  x <- fresh
  check e x
  pure (e, x)
```

I've moved over both expression and pattern type inference (the two which used to suck the most) to the new bidirectional system, but kinds are pretty much the same.

### DIFF
--- a/src/Control/Monad/Infer.hs
+++ b/src/Control/Monad/Infer.hs
@@ -6,7 +6,7 @@
   , StandaloneDeriving
   , MultiParamTypeClasses
   , OverloadedStrings
-  , ConstraintKinds#-}
+  #-}
 module Control.Monad.Infer
   ( module M
   , TypeError(..)
--- a/src/Control/Monad/Infer.hs
+++ b/src/Control/Monad/Infer.hs
@@ -5,17 +5,19 @@
   , ConstraintKinds
   , StandaloneDeriving
   , MultiParamTypeClasses
-  , OverloadedStrings #-}
+  , OverloadedStrings
+  , ConstraintKinds#-}
 module Control.Monad.Infer
   ( module M
   , TypeError(..)
   , Constraint(..)
   , Env(..)
   , MonadInfer
-  , lookupTy, lookupTy', fresh, freshFrom, runInfer, extend
+  , lookupTy, lookupTy', fresh, freshFrom, runInfer, extend, freeInScope
   , extendKind, extendMany, extendManyK
   , difference, freshTV, freshKV
   , instantiate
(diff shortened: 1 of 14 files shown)


src/Control/Monad/Infer.hs
```diff
@@ -53,9 +55,16 @@ instance Semigroup Env where
...
 data Constraint p
-  = ConUnify (Expr p) (Type p) (Type p)
+  = ConUnify SomeReason (Type p) (Type p)
+  | ConSubsume SomeReason (Type p) (Type p)
 deriving instance (Show (Ann p), Show (Var p), Show (Expr p), Show (Type p))
   => Show (Constraint p)
+deriving instance (Eq (Ann p), Eq (Var p), Eq (Expr p), Eq (Type p))
+  => Eq (Constraint p)
+
+instance Eq (Constraint p) => Ord (Constraint p) where
```
SquidDev: Are you sure this implementation is acceptable?

zardyh: It's not "acceptable", but it's kinda needed: We have to solve unification constraints before subsumption ones otherwise stuff will go bad. 

Stuff is still currently going bad, but it'll go badder.

zardyh:
I'm still working on some things, actually. Somehow we type _less_ programs now.

zardyh:
The latest commit makes one change that allows for _some_ level of higher-rankedness: Variables _checked_ against a polymorphic type use the subsumption relationship, while variables synthesising their type use the previous instancing behaviour.

We could previously type this program:

```ocaml
let higher_rank (f : forall 'a. 'a -> 'a) =
  (f 1, f "foo")
(* higher_rank : (∀ 'a. 'a -> 'a) -> int * string
```

We can now type _some_ uses of this function. Namely:
```ocaml
let foo =
  let id : forall 'a. 'a -> 'a =
    fun x -> x
  in higher_rank_id id ;;
(* main : int * string *)

let id x = x
and bar = higher_rank id ;;
(* bar : int * string *)
```

but not:
```ocaml
higher_rank (fun x -> x)
```

as functions never synthesise a polymorphic type.

PR #3, opened 2018-02-06
WIP: Happy parsing

SquidDev:
Because the current one makes me sad.

 - [x] Types
 - [x] Top level expressions
 - [x] Expression
   - [x] Record syntax
   - [x] Operators
   - [x] Operator sections
   - [x] Record access


SquidDev:
We're gonna want to squash this, as it's a total mess. But it (seems to) work.

zardyh:
Can you change the names back to `Parser.Lexer` and `Parser`?

amuletml.cabal
```diff
@@ -14,14 +14,18 @@ executable amuletml
...
+                     , bytestring >= 0.10.8.2 && < 0.11
+                     , base >= 4.9 && < 4.11
                      , lens >= 4.15 && < 4.16
                      , parsec >= 3.1 && < 3.2
                      , monad-gen >= 0.3 && < 0.4
                      , containers >= 0.5 && < 0.6
                      , transformers >= 0.5 && < 0.6
                      , pretty-show
+  build-tools:         happy, alex
+  other-modules:       Parser.HParser, Parser.ALexer
```
zardyh: I would prefer if these names were less silly. Have you removed the Parsec ones? Can you kill the Parsec dependency?

default.nix
```diff
@@ -7,14 +7,18 @@ let
   f = { mkDerivation, stdenv
       , mtl
       , syb
-      , base
       , text
+      , array
+      , bytestring
+      , base
+      , lens
```
zardyh: Do you have Awk? If so,
```
!awk '{ print length, $0 }' | sort -n | cut -d' ' -f2-
```
from visual mode.

zardyh: (otherwise, sort them by length, but keep mkDerivation and stdenv at the top)

src/Parser/ALexer.x
```diff
@@ -0,0 +1,203 @@
...
+  <0> $lower $ident*                   { lexTok TcIdentifier }
+  <0> $upper $ident*                   { lexTok TcConIdent }
+  <0> \_ $ident+                       { lexTok TcHole }
+  <0> \. $ident+                       { lexTok (TcAccess . T.tail) }
+  <0> \' $lower $ident*                { lexTok (TcTyVar . T.tail) }
+  <0> \"                               { beginString }
+
+  <string> \" { endString }
+
+  <string> \\ a { appendChar '\a' }
```
zardyh: Nice use of startcodes.

src/Parser/AWrapper.hs
```diff
@@ -0,0 +1,154 @@
...
+  show (POK _ s) = show s
+  show (PFailed msg pos) = show pos ++ ": " ++ msg
+
+instance Pretty a => Pretty (ParseResult a) where
+  pprint (POK _ s) = pprint s
+  pprint (PFailed msg pos) = pos <+> ": " <+> msg
+
+newtype Parser a = P { unP :: PState -> ParseResult a }
+
+instance Functor Parser where
```
zardyh: These instances make me deeply unhappy.

src/Parser/AWrapper.hs
```diff
@@ -0,0 +1,154 @@
...
+  fmap = liftM
+
+instance Applicative Parser where
+  pure a = a `seq` (P $ flip POK a)
+  (<*>) = ap
+
+instance Monad Parser where
+  (P m) >>= k = P $ \s -> case m s of
+                            POK s' a -> unP (k a) s'
```
zardyh: Can you indent this by two spaces instead of having it hanging?

src/Parser/HParser.y
```diff
@@ -0,0 +1,299 @@
...
+     | '(' List(Pattern, ',') ')' { withPos2 $1 $3 $ tuplePattern $2 }
+     | ArgP ':' Type              { withPos2 $1 $2 $ PType $1 $3 }
+
+Arm :: { (Pattern Parsed, Expr Parsed) }
+    : '|' Pattern '->' Expr       { ($2, $4) }
+
+
+Type :: { Type Parsed }
+     : TypeProd                                   { $1 }
+     | TypeProd '->' Type                         { TyArr $1 $3 }
```
zardyh: Does this mean `a * b -> c` is `(a * b) -> c` or `a * (b -> c)`?

src/Parser/HParser.y
```diff
@@ -0,0 +1,299 @@
...
+
+TyVar :: { Var Parsed }
+      : tyvar { Name (getIdent $1) }
+
+BindGroup :: { [(Var Parsed, Expr Parsed, Ann Parsed)] }
+          : Binding                           { [$1] }
+          | BindGroup and Binding             { $3 : $1 }
+
+Binding :: { (Var Parsed, Expr Parsed, Ann Parsed) }
+        : Var ListE(ArgP) '=' Expr            { (getL $1, foldr (\x y -> withPos2 x $4 (Fun x y)) $4 $2, withPos2 $1 $4 id) }
```
zardyh: Hmm, seems like you forgot the optional type annotation here:
```
| Var ListE(ArgP) ':' Type '=' Expr = { Ascription {- the stuff -} {- the type -} }
```

default.nix
SquidDev: I dunno, I'd prefer if they were alphabetical.

src/Parser/AWrapper.hs
SquidDev: Why?

default.nix
zardyh: They're sorted by length in the cabal file :< I also sort my imports by length, though in reverse order.

zardyh:
This one is definitely getting squashed.

src/Parser/HParser.y
SquidDev: `(a * b) -> c` IIRC. I'll double check though.

src/Parser/HParser.y
SquidDev: Done.

src/Parser/AWrapper.hs
zardyh: Defining Functor instances in terms of the Monad instance is reverse what you should be doing :P

zardyh:
LGTM™.

PR #4, opened 2018-02-10
Irrefutable Record Pattern Lowering

zardyh:
Record patterns are lowered into a series of nested pattern matches on
record accessors corresponding to the original.

```ocaml   
match foo with
| { x = k, y = l } -> ...
```    
=> 
```ocaml
match foo.x with
| k -> match foo.y with
  | l -> ...
```

This means that the optimiser has more stuff do to but the Core grammar is less crap.


zardyh:
This is so behind I'm not even going to _try_ merging it.

PR #5, opened 2018-02-14
Modules

SquidDev:
I guess??


src/Parser.y
```diff
@@ -99,15 +104,19 @@ Tops :: { [Toplevel Parsed] }
...
+    | external val ident':' Type '=' string    { withPos2 $1 $7 $ ForeignVal (getName $3) (getString $7) $5 }
 
-    | type Var ListE(TyVar)                          { TypeDecl (getL $2) $3 [] }
-    | type Var ListE(TyVar) '=' List1(Ctor, '|')     { TypeDecl (getL $2) $3 $5 }
-    | type Var ListE(TyVar) '=' '|' List1(Ctor, '|') { TypeDecl (getL $2) $3 $6 }
+    | type ident ListE(TyVar)                          { TypeDecl (getName $2) $3 [] }
+    | type ident ListE(TyVar) '=' List1(Ctor, '|')     { TypeDecl (getName $2) $3 $5 }
+    | type ident ListE(TyVar) '=' '|' List1(Ctor, '|') { TypeDecl (getName $2) $3 $6 }
+
+    | module Con '=' Tops end                  { Module (getL $2) $4 }
```
SquidDev: `module Con '=' begin Tops end`

examples/modules.ml
```diff
@@ -0,0 +1,11 @@
+module Option =
```
zardyh: Can you make this have an explicit `begin` token? I'm fine with either
```ocaml
module Option = struct
...
end
```
or
```ocaml
module Option = begin
...
end
```

src/Core/Lower.hs
```diff
@@ -203,16 +203,20 @@ lowerPat pat = case pat of
...
+                                 (k, _) <- makeBigLams ant
+                                 (v,,) <$> lowerType ant <*> (k <$> lowerExpr ex)))
+      <*> lowerProg prg
+lowerProg (TypeDecl (TvName var) _ cons:prg) =
+  (:) <$> (CosType var <$> do
+              for cons $ \case
+                UnitCon (TvName p) (_, t) -> (,) p <$> lowerType t
+                ArgCon (TvName p) _ (_, t) -> (,) p <$> lowerType t)
+      <*> lowerProg prg
+lowerProg (Open _ _:prg) = lowerProg prg
```
zardyh: Idea: lower this to a bunch of lets accessing records, I think.

```ocaml
open (Foo : module val foo; val bar end)
(* into *)
let foo = Foo.foo
let bar = Foo.bar
```

src/Core/Lower.hs
```diff
@@ -203,16 +203,20 @@ lowerPat pat = case pat of
...
+lowerProg (Module _ b:prg) = (++) <$> lowerProg b <*> lowerProg prg
```
zardyh: Here, float out the types and make the values into a record.

src/Syntax.hs
```diff
@@ -26,15 +26,26 @@ newtype Typed = Typed Typed deriving Data
 
 data family Var a
 
+class InModule v where
+  inModule :: v -> v -> v
```
zardyh: Can you not use `Monoid` for this - or, well, `Semigroup`?

SquidDev: I'm 90% sure I can replace this with an empty space. I don't think I ended up using it at all.

zardyh:
This looks good, but the fact that modules don't have an explicit computational interpretation makes me wary for the future - I'm thinking of things like first class modules and implicit module parameters, which aiui modules as implemented like this don't let us do.

src/Syntax.hs
zardyh: Oh.

src/Syntax/Resolve.hs
```diff
@@ -93,17 +93,7 @@ resolveModule (r:rs) = flip catchError (throwError . flip ArisingFromTop r)
...
-           modP (Name v) = InModule v
-           modP (InModule m v) = InModule m . modP v
-
-           modR _ x@(TgInternal _) = x
-           modR m (TgName v i) = TgName (T.intercalate (T.pack ".") (modParts m [v])) i
-
-           modParts (Name v) xs = v:xs
-           modParts (InModule m v) xs = m:modParts v xs
-
+           modZip name name' v v' = zip (map (name<>) v) (map (name'<>) v')
```
zardyh: blessed diff, :+1: squid

zardyh:
Is this done, @SquidDev?

src/Parser.y
zardyh: done

zardyh:
Is this done now?

SquidDev:
Yes?

PR #6, opened 2018-02-15
Skolem Type Constants

zardyh:
I'm not sure if these are sound, but cursory inspection at least tells me they're half logical.

I'm requesting a review from myself as well since it's past midnight and I have no idea if this code actually works or not.

### DIFF
--- a/src/Core/Optimise/Inline.hs
+++ b/src/Core/Optimise/Inline.hs
@@ -42,6 +42,8 @@ betaReduce = pass go where
       pure . CotLet [(var', tp, ex)] .  substitute (Map.singleton var ref) $ body
     CotTyApp (CotLam Big (var, _) body) tp ->
       pure $ substituteInTys (Map.singleton var tp) body
+    CotTyApp body@CotLam{} tp -> pure body -- TODO: This really shouldn't happen but it does
+    -- We need it to generate better code in the rank-N case
     _ -> pure term
 
 score :: CoTerm -> Trans Int
--- a/src/Core/Optimise/Inline.hs
+++ b/src/Core/Optimise/Inline.hs
@@ -42,7 +42,7 @@ betaReduce = pass go where
       pure . CotLet [(var', tp, ex)] .  substitute (Map.singleton var ref) $ body
     CotTyApp (CotLam Big (var, _) body) tp ->
       pure $ substituteInTys (Map.singleton var tp) body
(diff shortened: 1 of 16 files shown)


zardyh:
apparently I can't request a review from myself :(

src/Types/Unify.hs
```diff
@@ -36,6 +36,12 @@ bind var ty
 unify :: Type Typed -> Type Typed -> SolveM ()
 unify (TyVar a) b = bind a b
 unify a (TyVar b) = bind b a
+
+unify t@TySkol{} b
```
SquidDev: Why this rather than
```haskell
unify TySkol{} TySkol{} = pure ()
unify t@TySkol{} b = throwError $ SkolBinding t b
unify b t@TySkol{} = throwError $ SkolBinding t b
```

I know it's more cases, I guess I just find it cleaner.

src/Core/Optimise/Inline.hs
```diff
@@ -42,6 +42,8 @@ betaReduce = pass go where
       pure . CotLet [(var', tp, ex)] .  substitute (Map.singleton var ref) $ body
     CotTyApp (CotLam Big (var, _) body) tp ->
       pure $ substituteInTys (Map.singleton var tp) body
+    CotTyApp body@CotLam{} _ -> pure body -- TODO: This really shouldn't happen but it does
```
SquidDev: Doesn't this just strip all `CotTyApp`s? Not that I'm complaining, just trying to understand the rationale.

zardyh: Uh, yes, it was generating broken code and removing inlining opportunities, and I didn't like that. It's my fault (in the type checker somewhere, probably). Feel free to remove it.

zardyh: Actually, I'm not sure we even need this anymore.

zardyh: We _do_ - and I'll be glad to remove it along with all the other types from Core. 
They're far more trouble than they're worth.

Without,
```
let const x _ = 0 ;;
let rankn (x : forall 'a. 'a -> int) = (x 1, x ());;
let main = rankn (const 0)
```
becomes
```
let { 
  main : { 1 : int, 2 : int } = 
    { {} | 1 : int = (λ(ab : w). 0 @int) (1), 2 : int = (λ(ab : w). 0 @unit) (unit) }
 }
```

which is just abhorrent.

PR #8, opened 2018-02-20
Improve Pretty

SquidDev:
This means we no longer have to hide imports


src/Pretty.hs
```diff
@@ -42,6 +52,13 @@ data Style
   | Operator
   deriving (Eq, Show, Ord)
 
+infixr 5 <\>,<\\>
+
+-- Alias various definitions so we're not masking existing monads
```
zardyh: masking existing functions*

PR #9, opened 2018-02-21
Detect ambiguous references in the resolver

SquidDev:
Because I'm incapable of actually working on what's needed.

### DIFF
--- a/src/Errors.hs
+++ b/src/Errors.hs
@@ -74,6 +74,7 @@ instance Pretty TypeError where
 
 instance Pretty ResolveError where
   pretty (R.NotInScope e) = string "Variable not in scope:" <> verbatim e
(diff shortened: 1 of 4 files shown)


src/Errors.hs
```diff
@@ -74,6 +74,7 @@ instance Pretty TypeError where
 
 instance Pretty ResolveError where
   pretty (R.NotInScope e) = string "Variable not in scope:" <> verbatim e
+  pretty (R.Ambiguous v _) = string "Ambiguous reference to variable:" <> verbatim v
```
zardyh: these aren't going to be printed with spaces, you want `<+>`

zardyh:
This is deathly silly—I'd be fine with a warning/error for repeated bindings within the same binding group, but this isn't ambiguous at all.

PR #10, opened 2018-02-21
Use type-directed lowering

zardyh:
Instead of guessing when things should be quantified, we can use the
fact that we statically know the types of everything to figure out
precisely when things should be quantified.

This does that.

### DIFF
--- a/src/Core/Lower.hs
+++ b/src/Core/Lower.hs
@@ -15,9 +15,7 @@ import Types.Infer (tyString, tyInt, tyBool, tyUnit)
 
 import qualified Data.Text as T
 import Data.Traversable
-import Data.Function
 import Data.Span
-import Data.List
 
 import Core.Core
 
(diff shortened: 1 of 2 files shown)


src/Core/Lower.hs
```diff
@@ -50,101 +48,78 @@ errRef = CotRef (TgInternal "error")
...
+lowerAnyway :: MonadLower m => Expr Typed -> m CoTerm
+lowerAnyway (Record xs _) = case xs of
+  [] -> pure (CotLit ColRecNil)
+  xs -> do
+    xs' <- for xs $ \(label, ex) ->
+      (,,) <$> pure label <*> lowerType (getType ex) <*> lowerExpr ex
+    pure (CotExtend (CotLit ColRecNil) xs')
+lowerAnyway (RecordExt e xs _) = do
+  xs' <- for xs $ \(label, ex) ->
+    (,,) <$> pure label <*> lowerType (getType ex) <*> lowerExpr ex
```
SquidDev: Tuple sections? I know you don't like them, so fair doos if not.

zardyh: Eh, sure.

zardyh:
I need to see if this lets me remove my _favourite_ hack from the compiler, [this](https://github.com/zardyh/amulet/pull/6#discussion_r168627743)

zardyh:
It does!

zardyh:
This is done.

PR #12, opened 2018-02-23
Remove CotBegin

zardyh:
Pretty self-explanatory. `Begin` desugars into a flat `let`. I think this lets us kill pure expressions in `begin` with the same logic as for `let`s. Don't quote me on that.


zardyh:
I did _not_ intend for this to be of this size.

PR #13, opened 2018-02-27
Hell yeah, it's ANF

SquidDev:
It's like normal normal form, but administrative instead.


zardyh:
That's a lot of commits.

src/Core/Core.hs
```diff
@@ -152,31 +147,26 @@ instance Pretty a => Pretty (CoStmt a) where
 instance Pretty a => Pretty [CoStmt a] where
   pretty = vcat . map pretty
 
-{-# ANN freeIn "HLint: ignore" #-}
```
zardyh: Bonus points for removing this

src/Core/Lower.hs
```diff
@@ -110,67 +139,60 @@ lowerAt (Access r k _) ty = do
...
-tup2Rec k b = do
-  b' <- lowerType b
-  pure [(T.pack (show k), b')]
+lowerType :: Type Typed -> CoType
+lowerType t@TyTuple{} = CotyExactRows (tup2Rec 1 t)
+lowerType (TyArr a b) = CotyArr (lowerType a) (lowerType b)
+lowerType (TyForall vs b) = foldr (CotyForall . unTvName) (lowerType b) vs
+lowerType (TyApp a b) = CotyApp (lowerType a) (lowerType b)
+lowerType (TyRows rho vs) = CotyRows (lowerType rho) (map (fmap lowerType) vs)
+lowerType (TyExactRows vs) = CotyExactRows (map (fmap lowerType) vs)
```
zardyh: `map (fmap` must be some kind of sin. I'll allow it, though.

src/Core/Optimise/Inline.hs
```diff
@@ -1,103 +1,84 @@
...
-betaReduce :: TransformPass
-betaReduce = pass go where
-  go term = case term of
-    CotApp (CotLam Small (var, tp) body) ex -> do
-      (var', ref) <- invent tp
-      pure . CotLet [(var', tp, ex)] .  substitute (Map.singleton var ref) $ body
-    CotTyApp (CotLam Big (var, _) body) tp ->
-      pure $ substituteInTys (Map.singleton var tp) body
-    _ -> pure term
+inlineVariablePass :: IsVar a => [CoStmt a] -> [CoStmt a]
```
zardyh: Is this still live?

src/Core/Simplify.hs
```diff
@@ -2,33 +2,26 @@ module Core.Simplify
...
-      , matchKnownConstr
-      , matchOfMatch
-      , matchOfBottom
-      , inlineVariable
-      , betaReduce
-      , dropUselessLet
-      -- , complexSafePropag
-      ]
+optmOnce = pure . passes where
+  passes = foldl (.) id $ reverse
```
zardyh: This is `mconcat` for the Endo monoid

src/Core/Optimise/Inline.hs
SquidDev: Yeah, this now works again. It's horribly slow in GHCi, but it works.

PR #14, opened 2018-02-28
Some fancy dead code analysis

SquidDev:
This strips any let binding whose variable is not used in any live expression and whose value is pure. What qualifies as pure I hear you ask?

 - Atoms (references, lambdas, literals)
 - Partially applied functions
 - Constructors
 - Match expressions with one case (and no bindings are consumed)
 - Binary operators

### DIFF
--- a/src/Core/Optimise/DeadCode.hs
+++ b/src/Core/Optimise/DeadCode.hs
@@ -24,10 +24,11 @@ deadCodePass = snd . freeS (DeadScope mempty) Nothing where
   freeS _ (Just m) [] = (VarSet.singleton (toVar m), mempty)
   freeS _ Nothing  [] = (mempty, mempty)
 
-  freeS s m (x@(CosForeign v _ _):xs) = let (fxs, xs') = freeS s m xs
-                                    in if toVar v `VarSet.member` fxs
-                                       then (toVar v `VarSet.delete` fxs, x:xs')
-                                       else (fxs, xs')
+  freeS s m (x@(CosForeign v _ _):xs) =
+    let (fxs, xs') = freeS s m xs
+     in if toVar v `VarSet.member` fxs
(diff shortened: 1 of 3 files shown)


zardyh:
> Match expressions with one case (and no bindings are consumed)

I dun get it.

src/Core/Optimise/DeadCode.hs
```diff
@@ -24,10 +24,11 @@ deadCodePass = snd . freeS (DeadScope mempty) Nothing where
   freeS _ (Just m) [] = (VarSet.singleton (toVar m), mempty)
   freeS _ Nothing  [] = (mempty, mempty)
 
-  freeS s m (x@(CosForeign v _ _):xs) = let (fxs, xs') = freeS s m xs
```
SquidDev: You're wrong, but OK.

zardyh:
Ｎｏｔ　ｈａｐｐｙ　ａｂｏｕｔ　ｔｈａｔ　ｍｏｄ− ahem, sorry. Not happy about that module name.

src/Core/Optimise/DeadCode.hs
```diff
@@ -0,0 +1,155 @@
...
+  freeS _ (Just m) [] = (VarSet.singleton (toVar m), mempty)
+  freeS _ Nothing  [] = (mempty, mempty)
+
+  freeS s m (x@(CosForeign v _ _):xs) =
+    let (fxs, xs') = freeS s m xs
+     in if toVar v `VarSet.member` fxs
+           then (toVar v `VarSet.delete` fxs, x:xs')
+           else (fxs, xs')
+  freeS s m (CosLet vs:xs) =
+    let m' = find (\x -> case toVar x of
```
zardyh: Are you wrong here? :thinking:

Issue #11, opened 2018-02-22
Coercions and newtypes


zardyh:
It'd be lovely to have zero-cost "wrappers" around other types that get compiled way. For instance, the type declared below leads to the generation of both the constructor reference `ChurchList` and, wherever it is destructured, a full `match` (`error` case included).

```ocaml
type church 'a = ChurchList of forall 'r. ('a -> 'r -> 'r) -> 'r -> 'r ;;
```

This could be better if we added a new constructor to both `CoTerm` and `CoPattern`: `CotCast` and `CopCast`, respectively. A `Cast` constructor represents, at compile time, changing the type of a value from `a` to `b` provided they are representationally equal.

For instance, the type above introduces a statement of representational equality: `church 'a ~ forall 'r. ('a -> 'r -> 'r) -> 'r -> 'r`. Using this, fully saturated constructor applications of `ChurchList`, and pattern matches against it, can be rewritten to be operationally the identity function.

Of course, using "newtype constructors" in a higher-order fashion would mean still having a computational interpretation for them: a wrapper around a cast, that gets compiled to, again, the identity function.

```
let {
  ChurchList : forall 'a. (forall 'r. ('a -> 'r -> 'r) -> 'r -> 'r) -> church 'a =
    Λ ('a : *). λ (x : forall 'r. ('a -> 'r -> 'r) -> 'r -> 'r) = cast x
}
```

Casts would be stripped by the backend in the same fashion as type applications and big-lambda abstractions.

PR #18, opened 2018-03-05
Feature: Coercions

zardyh:
This PR adds terms of the form `x |> phi` where x is an atom and `phi` is a _coercion_. Coercions are a zero-cost way to witness an isomorphism between two types at compile time (which, (un)fortunately, the users can't exploit. yet.)

Coercions have one of the following forms:

- `a ~ b` relates the types `a` and `b`
- `dom phi`, where phi relates `a -> x` and `b -> y`, relates `a` and `b`
- `cod phi`, where phi relates `a -> x` and `b -> y`, relates `x` and `y`

Types of the form

```ocaml
type foo = Foo of a
```

become

```ocaml
type foo ...

let Foo x = (x |> a ~ foo)
```

Similarly, matching becomes matching on casts.

### DIFF
--- a/src/Backend/Compile.hs
+++ b/src/Backend/Compile.hs
@@ -144,6 +144,7 @@ compileTerm (App f e) = do
     (fl', _) -> LuaCall fl' [e']
 
 compileTerm (TyApp f _) = compileAtom f
+compileTerm (Cast f _) = compileAtom f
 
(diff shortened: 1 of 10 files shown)


zardyh:
Neglected to mention this closes #11 once the spooky action calms down.

PR #21, opened 2018-03-09
Generalised Algebraic Data Types

zardyh:
I decided to see if I could do this in under an hour. I took 73 minutes.
Calling that a success.

This adds a new form of constructor, the "generalised" constructor:

    | Foo : t

`t` must "end" in an application of the type being declared, i.e.

- forall vs. t where t is kosher
- a -> t where t is kosher
- an application of the type being declared

Pattern matching on a GADT constructor introduces local equalities /when
checking/. This lets us guarantee principal types, I think. The details
are fuzzy.

Matching on a GeneralisedCon that binds constraints 'cs' will wrap the
body's generated constraints in an implication:

    type term 'a =
      | Lit : 'a ~ int => int -> term 'a
      | Unit : 'a ~ unit => term 'a
      | Fun : forall 'c 'b. 'a ~ 'c -> 'b => ('c -> 'b) -> term 'a
      | App : forall 'c 'b. 'a ~ 'b => term ('c -> 'b) * term 'c -> term 'a
    ;;

    let eval (x : term 'a) : 'a =
      match x with
      | Lit l -> l

This is ok because the constraints end up looking like `a ~ int => a ~ int`
which is trivial. type variables mentioned in the left-hand-side of an
implication constraint (the "before") are not kept in the resulting
substitution.


zardyh:
@SquidDev Can you make GADT constructors be checked in fresh scopes? Having to use an explicit forall is a bit annoying.

zardyh:
Er, resolved.

examples/gadt.ml
```diff
@@ -1,7 +1,7 @@
 type term 'a =
   | Lit : int -> term int
-  | Fun : forall 'a 'b. ('a -> 'b) -> term ('a -> 'b)
-  | App : forall 'a 'b. term ('a -> 'b) * term 'a -> term 'b
+  | Fun : ('a -> 'b) -> term ('a -> 'b)
+  | App : term ('a -> 'b) * term 'a -> term 'b
```
zardyh: Lovely, thanks @SquidDev!

zardyh:
This is not so WIP anymore. @SquidDev are we OK to merge?

PR #22, opened 2018-03-10
Feature: Floating point numbers

zardyh:
Clearly less important than GADTs.


src/Backend/Compile.hs
```diff
@@ -305,20 +306,21 @@ compileMatch ex ps = EC $ \xs next -> (genIf next, xs)
...
-                 , ("^", "..")
-                 , ("<", "<")
-                 , (">", ">")
-                 , (">=", ">=")
-                 , ("<=", "<=")
-                 , ("==", "==")
-                 , ("<>", "~=")
-                 , ("||", "or")
-                 , ("&&", "and") ]
+ops = Map.fromList
```
SquidDev: Boooo, hiss :p

zardyh: You are wrong.

PR #23, opened 2018-03-11
Refactor Core bindings

zardyh:
Why did we do this again? Oh, right, join points.


src/Backend/Compile.hs
```diff
@@ -168,7 +168,7 @@ compileTerm (Extend tbl exs) = do
 
         compileRow (f, _, e) es = (:es) . LuaAssign [LuaIndex (LuaRef new) (LuaString f)] . pure <$> compileAtom e
 
-compileTerm (Let [(x, _, e)] body)
+compileTerm (Let (One (x, _, e)) body)
   | usedWhen x == 1 && not (isMultiMatch e) && not (occursInTerm x e) = do
```
SquidDev: We should be able to remove the `not (occursInTerm x e)` check now.

PR #24, opened 2018-03-13
Add backtick support to the lexer/parser

SquidDev:
Pretty simple. Currently acts as a left-associative operator which binds more tightly than anything else. We can change later if it would make more sense.


zardyh:
Sorry for the delay, but LGTM.

PR #25, opened 2018-03-14
Add as-patterns

zardyh:
I, too, like easy things once in a while.


SquidDev:
```ml
match x with
| Just _ as y -> y
| y -> fmap f y
```
gets lowered into:
```ml
match x with
| y -> match y with
       | Just _ -> y
| y -> fmap f y
```
which is just `x`, which really isn't what we want (OK, in this case it actually is, but in the more general case it isn't).

I'm not sure of a way to do this without modifying core. I suspect to do this correctly we'll have to start using non-nested patterns, which probably requires join points to do efficiently.

zardyh:
Oh joy.

SquidDev:
Soon™

PR #31, opened 2018-03-17
Add implicit tuples in matches

zardyh:
Now, it's admissible to use a non-empty, comma-separated list of
expressions as both the scrutinee and pattern in a Match:

    match x, ..., y with
    | p, ..., p' -> ...

is equivalent to

    match (x, ..., y) with
    | (p, ..., p') -> ...


zardyh:
Oh, you.

PR #32, opened 2018-03-17
Add a rather* primitive multi-file support, old chap

SquidDev:
So my editor decides that "very" is a weasel word, so here we are.

### DIFF
--- a/compiler/Errors.hs
+++ b/compiler/Errors.hs
@@ -7,6 +7,7 @@ import qualified Data.Text as T
 import Data.Foldable
 import Data.Position
 import Data.Spanned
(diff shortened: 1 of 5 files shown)


zardyh:
I feel terrible saying this, but, surprisingly, it looks good to me. Are we sure it works, though?

SquidDev:
> Are we sure it works, though?

I tested a couple of basic multi-file "projects" and it all seemed to work. Not that that's a guarantee though.

SquidDev:
Too late now!

zardyh:
Ah, the Amulet way of doing things.

PR #37, opened 2018-03-20
Use a custom transformer instead of everywhere

SquidDev:
Oh boy is this ugly. But it's more efficient.


zardyh:
Did you profile?

PR #38, opened 2018-03-21
Core Lint

SquidDev:
I still need to do casts (and thus coercions) and patterns, but I'm slightly less sure about how to do them.


zardyh:
Eugh. This will need lots of fixing with Dependent Amulet :/

zardyh:
A TODO (@SquidDev): Record coercions are not working out properly. What do we do about those?

zardyh:
Removing the WIP status as the full optimiser pipeline preserves typing, and this is ready for merge.

zardyh:
@SquidDev Look good?

SquidDev:
Who else is looking forward to round 2 of this in a couple of weeks?

PR #39, opened 2018-03-25
Minor syntax enhancements

SquidDev:
Because I needed a break from core lint. Sorry. The changes better described in the commit messages - some of them may need more testing than others, as I'm not 100% sure they're "correct".


SquidDev:
Tuple sections appear to be broken in this PR, so I'd like to fix them before merging.

zardyh:
Well, you've broken `@@`, which was my favourite hack.

PR #41, opened 2018-03-30
Add literal patterns

SquidDev:
No major changes, but just lots of small ones across files.


src/Types/Infer/Pattern.hs
```diff
@@ -98,6 +98,9 @@ checkPattern pt@(PType p t ann) ty = do
   _ <- subsumes pt t' it
   _ <- unify pt ty t'
   pure (PType p' t' (ann, t'), binds, cs)
+checkPattern pt@(PLiteral l ann) ty = do
+  _ <- unify pt ty (litTy l)
```
zardyh: I think this should be in the inferPattern judgement instead, if you could.

zardyh:
Looks good. Is it sane?

SquidDev:
It hasn't crashed on me yet.

PR #42, opened 2018-04-02
Dependent Amulet - Phase 1

zardyh:
I've decided to do this _incrementally_. Today, we have the first batch
of things:

- Kinds no longer exist. Types describe themselves: we have, e.g. int :
  type, and type : set 1. (type is an alias for set 0, only used by the
  pretty printer.)

- `TyForall` and `TyArr` are no longer part of the type grammer: they
  were both replaced by a single binding form, `TyPi`, which represents
  the many shapes that a Π can take in a dependently typed language.

  - `TyPi (Anon _) _` corresponds to the old `TyArr`
  - `TyPi (Implicit _ _) _` corresponds to the old `TyForall`

  Note that both still exist as pattern synonyms.

- `TyForall` now has three "arguments": the variable, the body, and a new
  field, the kind- er, type- of the bound variable.

Note that this doesn't change much- it's mostly trivial refactoring to
support the proper cool features.


zardyh:
The build failure in 4a7eadc was just me forgetting the test suite mid-refactor, nothing big. I mean, er, "test suite".

@SquidDev Can you look into building the `examples` directory as part of CI?

PR #45, opened 2018-04-03
WIP: Full Dependent Types

zardyh:
This is, presently, really stupid, because I couldn't be arsed to write
an interpreter for Syntax at 22:00.

In the future (i.e., before this PR is merged):

  - Do that
  - Promote *everything*!

Right now, we can already type some dependently-typed programs, e.g.:

```ocaml
external val print : forall 'a. 'a -> unit = "print" ;;
external val string_of_int : int -> string = "tostring" ;;

type nat = Z | S of nat ;;

type vect 'n 'a =
  | Nil : vect Z 'a
  | Cons : 'a * vect 'k 'a -> vect (S 'k) 'a ;;

let replicate : forall 'a. ('x : nat) -> 'a -> vect 'x 'a = fun x e ->
  match x with
  | Z -> Nil
  | S k -> Cons (e, replicate k e) ;;

let ppr_vec (f : 'a -> string) (x : vect 'n 'a) : string =
  match x with
  | Nil -> "Nil"
  | Cons (a, b) -> "Cons (" ^ f a ^ ", " ^ ppr_vec f b ^ ")" ;;

let main x = print @@ ppr_vec string_of_int (replicate (S (S (S (S Z)))) 1)
```

This program prints `Cons (1, Cons (1, Cons (1, Cons (1, Nil))))` when
run. It does, however,  produce horribly broken Core - which now needs
to be entirely redone - and having to use a type declaration separate
from the function itself, apart from the hack needed to support
dependent matching in `match` (which I am eager to remove) makes me a
bit disappointed in our syntax.

I could envison something like:

```ocaml
fun replicate : forall 'a. ('x : nat) -> 'a -> vect 'x 'a
  | replicate Z elem = Nil
  | replicate (S k) elem = Cons (elem, replicate k elem)
```

but I have not yet tried to implement this. I don't think it'd be very
hard - in fact, it might even be easier.


zardyh:
Note that this depends on #44.

zardyh:
Note also that every single one of our examples continues to build even under dependent Amulet.

zardyh:
All the interesting, already-done bits have been merged. The rest can happen in a separate PR.

PR #48, opened 2018-04-23
Add support for an "open in" construct

SquidDev:
This evaluates the expression within the scope of a module. Some examples:

```ml
module X = begin
  let a = 1
  and b = 2
end ;;

let main = begin
  X.a; (* As before, just a module index *)
  X.(a, b) (* Constructs a tuple of (X.a, X.b) *)
  X.(a + b) (* Evaluates X.a + X.b *)
  X.{ a = a } (* Evaluates { a = X.a } *)
end
```

This also changes the syntax for "import as" from `open X as Y` to `module Y = X`.


zardyh:
Looks great, thanks!

PR #50, opened 2018-05-06
Use a specialised variable type for core

SquidDev:
This allows us to attach additional annotations to variables in the future. Namely the arity of join points.

Variables are also tagged with their type (type, tyvar, value, coercion), which we might want to verify in core at a later date - not currently a priority however.

### DIFF
--- a/src/Core/Var.hs
+++ b/src/Core/Var.hs
@@ -23,9 +23,10 @@ instance Ord CoVar where
 
 data VarInfo
   = ValueVar
+  | DataConVar
+  | TypeConVar
   | TypeVar
-  | TyvarVar
-  | CoeVar
+  | CastVar
   deriving (Eq, Show, Ord, Generic, Data)
 
 makeLenses ''CoVar
@@ -34,3 +35,11 @@ makePrisms ''VarInfo
 
 instance Pretty CoVar where
   pretty (CoVar i v _) = text v <> scomment (string "#" <> string (show i))
+
+class (Eq a, Ord a, Pretty a, Show a) => IsVar a where
+  toVar :: a -> CoVar
+  fromVar :: CoVar -> a
+
+instance IsVar CoVar where
+  toVar = id
+  fromVar = id
--- a/src/Core/Var.hs
+++ b/src/Core/Var.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE DeriveGeneric, DeriveDataTypeable, TemplateHaskell #-}
+
+module Core.Var where
+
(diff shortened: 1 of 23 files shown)


src/Core/Var.hs
```diff
@@ -0,0 +1,36 @@
...
+
+instance Eq CoVar where
+  (CoVar a _ _) == (CoVar b _ _) = a == b
+
+instance Ord CoVar where
+  (CoVar a _ _) `compare` (CoVar b _ _) = a `compare` b
+
+data VarInfo
+  = ValueVar
+  | TypeVar
```
zardyh: this is a TypeConVar

src/Core/Var.hs
```diff
@@ -0,0 +1,36 @@
...
+  | TyvarVar
```
zardyh: This is a TypeVar

src/Core/Var.hs
```diff
@@ -0,0 +1,36 @@
...
+  | CoeVar
```
zardyh: This is a CastVar or CoercionVar but just use a word that exists

src/Core/Var.hs
```diff
@@ -0,0 +1,36 @@
...
+        }
+  deriving (Show, Generic, Data)
+
+instance Eq CoVar where
+  (CoVar a _ _) == (CoVar b _ _) = a == b
+
+instance Ord CoVar where
+  (CoVar a _ _) `compare` (CoVar b _ _) = a `compare` b
+
+data VarInfo
```
zardyh: Can I have a DataConVar? It'll probably be useful.

src/Data/VarSet.hs
```diff
@@ -50,21 +47,19 @@ union = coerce Set.union
...
-delete :: Var Resolved -> Set -> Set
-delete (TgName _ x) set = coerce (Set.delete x (coerce set))
-delete _ set = set
+delete :: CoVar -> Set -> Set
+delete (CoVar x _ _) set = coerce (Set.delete x (coerce set))
 
 isEmpty :: Set -> Bool
 isEmpty = coerce Set.null
 
 class (Data a, Eq a, Ord a, Pretty a, Show a) => IsVar a where
```
zardyh: This could be in Core.Var.
Also drop the `Data a` constraint if you can.

SquidDev: Yeah, I've been wanting to move it elsewhere for a while and wasn't sure where. This is as good a time as any!

PR #51, opened 2018-05-08
Indentation aware syntax

SquidDev:
Basically so we can write:

```ml
let f x = 
  match x with
  | a ->
      print "foo"
      print a
  | _ -> print "bar"
  print "baz"
```

This is heavily inspired by [F#'s handling of this](https://fsharp.org/specs/language-spec/4.0/FSharpSpec-4.0-latest.pdf "See page 279 and later").

 - [x] Support for all expressions (`let`, `match`, `begin`, `fun`, `function`)
 - [x] Ensure the current examples can be built with minimum modifications: we want as much backwards compatibility as possible.
 - ~~Produce warnings when there are dubious indentation changes.~~ I've done some work, but I'm postponing this as it's not an immediate priority and it's a little flaky.
 - [x] Ensure the virtual tokens don't show up in any of the error messages.

It might also be worth seeing where else we can eliminate shift-reduce conflicts in the parser. We've done a bit by requiring `end` after some expressions, but there's more to do.


SquidDev:
<3

PR #52, opened 2018-05-09
WIP: Comparison tests

SquidDev:
Based on the `tasty-golden` package, but with more fanciness. One defines a function `FilePath -> T.Text -> String` which consume file name and file contents, processes it (lex, parse, type check, etc...) and returns a string representation of the output. We then ensure the outputs are the same, producing a diff if not.

## Things we need to check
 - **Lexer**
   - [x] Unterminated constructs such as comments or strings. It might be worth improving the error messages to include the start position of these comments too.
   - [x] Basic tokens (literals, identifiers, etc...)
   - [ ] Context rules
   - [ ] Context warnings (I suspect this'll require some improvements to the warning system).
 - [ ] **Parser:** I'll expand this list when we get to it. I'm not sure what the passing case will be for this, obviously we can print the error in the failing case.
 - **Resolver:** Ideally we'll be able to change this to report multiple errors, as none of these are irrecoverable.
   - [x] Modules
   - [x] Ambiguous variables
 - [ ] **Typechecker:** I'm leaving this entirely in hydraz's hands.
 - [ ] **Lower, optimiser and codegen:** I don't think it's possible to split these up, but we could definitely check pre-optimiser and post-optimiser code.


SquidDev:
So I haven't done everything I wanted to do, but most of the "complex" things I wanted to test are, well, broken. So I'd like to get this merged, then fix the various broken components and add specific tests for them.

That being said, @zardyh if you've got any tests you want to add which pass (especially type checking ones), do commit them!

PR #53, opened 2018-05-13
Add support for lexing unicode

SquidDev:
Whilst Alex does support unicode strings, it ends up being rather hard to handle arbitrary character categories. Instead, we perform the same trick that GHC does: operate on latin1 encoded strings and map unicode categories to non-ASCII characters. This allows us to match entire blocks of characters with a single pattern.

We've also dropped our dependency on bytestring, as the lexer works on a lazy text value internally. This makes things a little cleaner, as we're no longer encoding and decoding the whole time.

The only downside to this is that it makes handling unicode keywords a little more ugly. See the `lexOperator` function in `Lexer.x` for instance.


zardyh:
This :clap: is :clap: great.

PR #54, opened 2018-05-17
Report multiple variable resolution errors

SquidDev:
This just replaces invalid terms with some "junk" value, adding the error to a writer.

It's a little ugly in places, and depends on `Control.Monad.Infer.Errors` - it might be worth moving this into a separate module when we generalise error reporting a little more.


SquidDev:
In my defence, hydraz unofficially approved it on Slack.

PR #55, opened 2018-05-18
Deep Instantiation

zardyh:
Basically if we have

    f : a -> forall 'b. ...

we generate evidence like

    λ (x : a). f x @α

where α is a fresh 'b.

See the new `rankn/deepinst01-pass.ml` for details.


src/Control/Monad/Infer.hs
```diff
@@ -162,12 +163,20 @@ instantiate tp@(TyPi (Implicit v _) ty) = do
     TgName n _ -> TgName n <$> gen
   let map = Map.singleton v var
 
-      squish f (Just x) = Just (x . f)
-      squish f _ = Just f
-
       appThisTy e = ExprWrapper (TypeApp var) e (annotation e, apply map ty)
   (k, _, t) <- instantiate (apply map ty)
   pure (squish appThisTy k, tp, t)
+instantiate tp@(TyPi (Anon co) od@dm) = do
```
zardyh: This is the only interesting thing, by the way.

zardyh:
Lots of noise because this changed how type variables are numbered.

src/Control/Monad/Infer.hs
```diff
@@ -282,3 +291,7 @@ prettyMotive :: SkolemMotive Typed -> Doc
 prettyMotive (ByAscription t) = string "of a type ascription" <#> string "against the type" <+> pretty t
 prettyMotive (BySubsumption t1 t2) = string "of a subsumption constraint relating" <+> pretty t1 <+> string "with" <+> pretty t2
 prettyMotive (ByExistential v t) = string "it is an existential" <> comma <#> string "bound by the type of" <+> pretty v <> comma <+> pretty t
+
+squish :: (a -> c) -> Maybe (c -> c) -> Maybe (a -> c)
+squish f (Just x) = Just (x . f)
```
SquidDev: `Just . maybe f (.f)` instead?

src/Syntax.hs
```diff
@@ -116,8 +116,21 @@ data Wrapper p
...
   | WrapVar (Var p) -- Unsolved wrapper variable
+  | WrapFn (WrapCont p)
   | IdWrap
 
+newtype WrapCont p = MkWrapCont { runWrapper :: Expr p -> Expr p }
+
+deriving instance Typeable p => Typeable (WrapCont p)
+instance Typeable p => Data (WrapCont p) where
+  gunfold _ _ = error "gunfold WrapCont"
+  toConstr _ = error "gunfold WrapCont"
```
SquidDev: Should probably be `toConstr` and `dataTypeOf`?

src/Control/Monad/Infer.hs
```diff
@@ -162,12 +163,20 @@ instantiate tp@(TyPi (Implicit v _) ty) = do
...
   pure (squish appThisTy k, tp, t)
+instantiate tp@(TyPi (Anon co) od@dm) = do
+  (wrap, _, dm) <- instantiate dm
+  let cont = fromMaybe id wrap
+  var <- fresh
+  let ty = TyPi (Anon co) dm
+      lam :: Expr Typed -> Expr Typed
+      lam e
+        | ann <- annotation e
+        = Fun (PType (Capture (TvName var) (ann, co)) co (ann, co)) (cont (App e (VarRef (TvName var) (ann, co)) (ann, od))) (ann, ty)
```
SquidDev: I assume this is the only way, but it seems to create a lot of noise: `2 + 3` expands to `(fun x -> fun y -> (+) x y) 2 3` for instance, which is a little un-ideal.

PR #56, opened 2018-05-18
Semantic pretty printing

SquidDev:
This separates the construction of "semantically annotated" documents from their rendering into ANSI strings. This brings several advantages:

 - Some operations are now generalised over all documents, not just `Doc Style` ones.
 - Our ANSI renderer can be smarter than the previous "concat everything together" one was. For instance, nested annotations now render as expected.

The other major change here is converting the pretty printer to produce `Text` instead of `String`s.

I apologise for the rather noisy PR - moving the package, fixing trailing whitespace and changing the bullet meant a lot of files have a single line diff.

### DIFF
--- a/amuletml.cabal
+++ b/amuletml.cabal
@@ -87,10 +87,12 @@ executable lexer
                      , Parser.Wrapper
                      , Parser.Unicode
 
+                     , Text.Pretty
+                     , Text.Pretty.Semantic
+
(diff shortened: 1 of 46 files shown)


zardyh:
Looks good! Is this ready?

SquidDev:
Yep.

PR #58, opened 2018-05-18
Generalised error reporting

SquidDev:
`src/Text/Pretty/Note.hs` is the main file of note (heh), as well as any implementations of the `Note` class.


zardyh:
Looks good to me.

PR #59, opened 2018-05-19
Feature: Visible Quantification

zardyh:
This PR adds a new form of type, `forall ('a : k) -> t`, which is to be used both visibly and relevantly. This is _not_ dependent types, yet.

The "regular" behaviour of `forall 'a. t` can be regained when you have a function of type `forall ('a : k) -> t` by using `f ?`.

We still don't have visible type lambdas, because that would frankly be quite a mess.


zardyh:
For extra reference: These types are still erased, and no changes were made to Core. :tada:

Issue #30, opened 2018-03-17
REPL


SquidDev:
We should have a REPL, with all the fancy things. 

 - [x] Multiline input: I don't know if we can detect EOF errors from the parser (but not the lexer) and thus prompt for another line.
 - [ ] Autocompletion
 - [x] Some fancy repl commands, like `:view`, `:reload`, etc... Though not strictly needed right now.

PR #60, opened 2018-05-20
WIP: REPL

SquidDev:
See #30. 
 - [x] Allow entering expressions as well as top-level statements. I've no clue how to handle this in a sane manner.
 - [x] Multiline input: I don't know if we can detect EOF errors from the parser (but not the lexer) and thus prompt for another line.
 - [ ] Autocompletion
 - [ ] Some fancy repl commands, namely `:view` (dump the definition) and `:type` (dump the type of an expression).
 - [x] Verify that we're using the Lua API correctly. I need to read up on the docs a little more, as I'm not sure we always pop from the stack correctly.


src/Backend/Lua.hs
```diff
@@ -14,4 +15,6 @@ import Core.Var
 import Syntax.Types
 
 compileProgram :: IsVar a => Env -> [Stmt a] -> LuaStmt
-compileProgram e = LuaDo . addOperators . emitProgram e . tagOccursVar
+compileProgram e = LuaDo . (unitDef :) . addOperators . emitProgram e . tagOccursVar where
+  unitDef = LuaLocal [ LuaName "__builtin_unit" ] [ LuaTable [ (LuaString "__tag", LuaString "__builtin_unit") ] ]
```
SquidDev: It might be nice to have some sort of `preamble` definition in `Postprocess`. This way we're not duplicating stuff between here and the REPL.

zardyh: Probably.

SquidDev:
Well, aside from me failing to merge it correctly, this works rather well. Thanks for the work on this ♡.

PR #65, opened 2018-06-03
Verify Post-TC

zardyh:
Currently only that recursive definitions are well-formed.

```
test.ml[17:5 ..17:9]: error
  Invalid recursive right-hand side for variable x
  • Note: because the variable x is not under a function
  Arising from use of the declaration
   │ 
17 │ let x = x
   │ 
```

The check is very silly: None of the variables bound in the same binding
group can appear outside of a function. Implicit and explicit laziness
counts as "a function".


compiler/Main.hs
```diff
@@ -72,14 +74,17 @@ compile (file:files) = runGen $ do
...
                 Right (prog, env') ->
-                  let (var, tys) = extractToplevels parsed
-                      (var', tys') = extractToplevels resolved
-                  in pure $ Right (tops ++ prog
-                                  , scope { RS.varScope = RS.insertN (RS.varScope scope) (zip var var')
-                                          , RS.tyScope  = RS.insertN (RS.tyScope scope)  (zip tys tys')
-                                          }
-                                  , modScope'
-                                  , env')
+                  case runVerify (verifyProgram prog) of
```
SquidDev: I wonder if we should look into shifting this into a monad error or something.

zardyh: We probably should, but that'll be Hard tm

zardyh: because they all have different error types

src/Data/Span.hs
```diff
@@ -63,3 +63,4 @@ instance Semigroup Span where
     | x == internal = y
     | y == internal = x
     | otherwise = error $ "<> spans of different files: " ++ fa ++ ", " ++ fb
+
```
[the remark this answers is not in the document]
zardyh: damn it sorry

compiler/Main.hs
SquidDev: I was thinking something like `` `catchError` (throwError . Verify)`, where `Verify` is some constructor for a union of all error types.

Or we do something with existentials :p.

zardyh: That won't work.

zardyh:
@SquidDev This is done.

PR #66, opened 2018-06-04
Cherry-picked join point conversion

plt-amy:
This is basically the same code as on the ~old~ _other_ branch: We
invent a ~join point~ _regular-ass lambda_ (which serves as a
continuation) when we see the result of a `match` being let-bound.

This breaks the match_consumed test and I'd like @SquidDev's opinion it.

### DIFF
--- a/tests/lua/match_consumed.ml
+++ b/tests/lua/match_consumed.ml
@@ -1,6 +1,7 @@
 (* This ensures that matches which are not in the tail position correctly
    pop all pending expressions on the stack and generate the appropriate
-   declaration before the join. *)
+   declaration before the join.
+   *)
 
 type option 't = None | Some of 't
 


zardyh:
Old:
```lua
  local function main (f)
    return function (x)
      local a = f(1.0)
      local b
      if x.__tag == "None" then
        b = 0.0
      elseif x.__tag == "Some" then
        b = x[1.0] * 2.0
      end
      return f(a + b)
    end
  end
```
Now:
```lua
 local function main (f)
    return function (x)
      local a = f(1.0)
      if x.__tag == "None" then
        return f(a)
      elseif x.__tag == "Some" then
        local cx = x[1.0] * 2.0
        local cu = _plus(a)
        return f(cu(cx))
      end
    end
  end
```

zardyh:
I think that expression ordering prevents us from currently turning that into `f(a + x[1] * 2)`, which is optimal (and also, kinda, what is represented).

PR #67, opened 2018-06-07
New-type elimination (again)

zardyh:
Newtype optimisation is back! Now with, hopefully, 100% less broken Core
Lints. I've added a small test about how newtypes are _compiled_ (this
is likely to break in the future if we ever manage to inline
constructors, by the way).

This optimisation is unreasonably efficient in Amulet, considering that
all types with a single constructor (and no existentials) are eligible
for newtype elimination, considering we have no curried constructors.
This includes single-constructor GADTs, since evidence parameters are
currently erased during Lower.

The idea is the same as before: we turn a type like

```ocaml
type foo 'a {
  Foo : forall ('a : *). 'a -> foo 'a
}
```

into a worker like

```ocaml
type foo 'a {}
let Foo : forall ('a : *). 'a -> foo 'a =
  Λ ('a : *). λ(x : 'a). x |> 'a ~ foo 'a
```

which will probably inline everywhere.

### DIFF
--- a/src/Core/Simplify.hs
+++ b/src/Core/Simplify.hs
@@ -4,7 +4,7 @@ module Core.Simplify
 
 import Data.List
 
--- import Core.Optimise.Newtype
+import Core.Optimise.Newtype
 import Core.Optimise.DeadCode
 import Core.Optimise.Inline
 import Core.Optimise.Reduce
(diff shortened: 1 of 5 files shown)


src/Core/Simplify.hs
```diff
@@ -21,15 +21,14 @@ lintPasses :: Bool
 lintPasses = True
 
 optmOnce :: [Stmt CoVar] -> Namey [Stmt CoVar]
-optmOnce = passes where
+optmOnce = passes <=< killNewtypePass where
```
SquidDev: No, evil!

PR #70, opened 2018-06-11
Don't run the reduce pass in the REPL

Lignum:
The existential example doesn't work in the REPL because of this:
```
> type box =
.   | Box : ('a -> string) * 'a -> box ;;

> let show_it (Box (f, x)) = f x
amc: src/Core/Optimise/Reduce.hs:384:9-50: Irrefutable pattern failed for pattern Just ty
```

@zardyh told me to do this to make Amulet seem like it has contributors.


zardyh:
I neither confirm nor deny these allegations.

SquidDev:
The comment is still right - we need to do _some_ commuting conversion in order to ensure codegen isn't trite

PR #71, opened 2018-06-11
Fix a pattern matching failure when using `force` in the REPL

Lignum:
Before:
```
> let pi = lazy (abs (atan (0.0 -. 1.0)) *. 4.0)
pi : lazy float =
 lazy <function>
> force pi
[string "do..."]:6: Pattern matching failure at =stdin[1:10 ..1:46] (ErrRun)
> 
```

After:
```
> let pi = lazy (abs (atan (0.0 -. 1.0)) *. 4.0)
pi : lazy float = lazy <function>
> force pi
_ : float = 3.141592653589793
> 
```


zardyh:
This is fallout from changing the representation of `()` away from `nil` and I'm rather sad that no tests caught it.
@squiddev runtime behavior tests when?

PR #73, opened 2018-06-25
Get rid of travis

SquidDev:


SquidDev:
Some context: Nix was resulting in Travis builds timing out (or taking incredibly long), as we needed to rebuild a reasonable amount of stuff, and it wasn't feasibly to selectively cache things.

PR #75, opened 2018-06-28
WIP: Kind annotations on type definitions

SquidDev:
@zardyh Help!


zardyh:
I ended up pushing something like this in the implicit parameters branch - so this PR is stale now. Sorry :/

PR #76, opened 2018-07-01
Feature: Implicit Parameters

zardyh:
Cool thing. Kinda like typeclasses, but without the coherence
requirement, though that might be enforced by the ambiguity check.
Local instances are available, since implicit value bindings can happen
at any let.

Possible choices for implicit parameters are stored in a trie keyed by
the spine of type applications, so a binding for `lazy int` would be
stored as

```haskell
Trie
  (fromList
     [ ( TyCon (TvName (TgName "lazy" (-34)))
       , Many
           (Trie
              (fromList
                 [ ( TyCon (TvName (TgInternal "int"))
                   , One
                       ImplChoice
                         { _implHead =
                             TyApp
                               (TyCon (TvName (TgName "lazy" (-34))))
                               (TyCon (TvName (TgInternal "int")))
                         , _implType =
                             TyApp
                               (TyCon (TvName (TgName "lazy" (-34))))
                               (TyCon (TvName (TgInternal "int")))
                         , _implPre = []
                         , _implVar = TvName (TgName "something" 1)
                         }
                   )
                 ]))
       )
     ])
```

Tuples, functions and records do not have unique prefixes, and so are
stored as-is.

The `implPre` field is used to record any extra steps that some choice
might entail. In particular:

- A choice might have other implicit parameters, as in
  `tests/implicits/implication.ml`. In this case, using that value
  entails solving its implicit parameters.
- A choice might be polymorphic. In this case, solving entails unifying
  the head type (`implHead`) with the wanted type for the implicit, and
  generating the correct type application.

The following is a test case which also serves as an example.

```ocaml
type monad 'm =
  Monad of (forall 'a. 'a -> 'm 'a)
         * (forall 'a 'b. 'm 'a -> ('a -> 'm 'b) -> 'm 'b)

type maybe 'a = Just of 'a | Nothing

let implicit monad_maybe =
  let bind x f =
    match x with
    | Just m -> f m
    | Nothing -> Nothing
  Monad (Just, bind)

let pure : forall 'm 'a. monad 'm => 'a -> 'm 'a =
  fun (Monad (f, _)) -> f

let (>>) : forall 'm 'a 'b. monad 'm => 'm 'a -> 'm 'b -> 'm 'b =
  fun _ x y -> x >>= fun _ -> y

let main = pure 1 >> Nothing
```

Blocking this merge:
- [x] Type inference for implicit parameters
  - [x] Implicit pattern syntax.
- [ ] More tests!


compiler/Test/Types/Check.hs
```diff
@@ -52,4 +52,5 @@ tests = do
   rankn <- testGroup "Rank-N inference tests" <$> goldenDir result "tests/rankn/" ".ml"
   visinst <- testGroup "Visible Instantiation inference tests" <$> goldenDir result "tests/visinst/" ".ml"
   lazy <- testGroup "Automatic laziness tests" <$> goldenDir result "tests/lazy/" ".ml"
-  pure (testGroup "Type inference" [ inference, gadts, rankn, visinst, lazy ])
+  implicit <- testGroup "Implicit parameter tests" <$> goldenDir result "tests/implicits/" ".ml"
```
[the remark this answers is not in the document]
zardyh: Done.

src/Core/Optimise/Reduce.hs
```diff
@@ -206,8 +206,8 @@ reduceTerm s (Cast (Ref v _) c)
   | Just (Cast a c') <- lookupVar s v
   , Just (l, r) <- relates c
   , Just (l', r') <- relates c'
-  , Just uni <- unify r l'
-  , Just _ <- unifyWith uni l r'
+  , True <- unifyClosed r l'
```
SquidDev: Why `True <- xyz` rather than just `xyz`?

src/Data/Reason.hs
```diff
@@ -52,6 +52,9 @@ instance (Data p, Data (Ann p), Data (Var p), Pretty (Var p)) => Reasonable Cons
 instance (Spanned (Ann p), Data p, Data (Ann p), Data (Var p), Pretty (Var p)) => Reasonable Toplevel p where
   blame _ = string "the" <+> highlight "declaration"
 
+instance (Spanned (Ann p), Pretty (Var p)) => Reasonable Binding p where
```
SquidDev: Not a required change, but I almost wonder if we should move some of these instances to `Syntax`.

src/Parser/Lexer.x
```diff
@@ -77,6 +77,7 @@ tokens :-
   <0> "end"    { constTok TcEnd }
   <0> "in"     { constTok TcIn }
   <0> "external" { constTok TcExternal }
+  <0> "implicit" { constTok TcImplicit }
```
SquidDev: I'm not convinced this is the nicest syntax, but eh.

src/Syntax/Implicits.hs
```diff
@@ -0,0 +1,232 @@
+{-# LANGUAGE StandaloneDeriving, FlexibleContexts, FlexibleInstances,
+   UndecidableInstances, ScopedTypeVariables, TemplateHaskell #-}
+module Syntax.Implicits where
```
SquidDev: Can we be a bit more selective on what we export?

src/Syntax/Resolve/Toplevel.hs
```diff
@@ -3,13 +3,12 @@ module Syntax.Resolve.Toplevel
...
 
-import Data.Triple
-
 import Syntax.Var
 import Syntax
 
 extractToplevel :: Semigroup (Var p) => Toplevel p -> ([Var p], [Var p])
-extractToplevel (LetStmt d) = (map fst3 d, [])
+extractToplevel (LetStmt d) = (map b d, []) where
+  b (Binding v _ _ _) = v
```
SquidDev: I'd like to say that this is a terrible method. Would it be worth turning `Binding` into a record?

src/Types/Infer/Builtin.hs
```diff
@@ -18,12 +18,13 @@ import Syntax
...
 import Text.Pretty.Semantic
 
-tyUnit, tyBool, tyInt, tyString, tyFloat, tyLazy :: Type Typed
+tyUnit, tyBool, tyInt, tyString, tyFloat, tyLazy, tyArrow :: Type Typed
 tyInt = TyCon (TvName (TgInternal "int"))
 tyString = TyCon (TvName (TgInternal "string"))
 tyBool = TyCon (TvName (TgInternal "bool"))
 tyUnit = TyCon (TvName (TgInternal "unit"))
 tyFloat = TyCon (TvName (TgInternal "float"))
+tyArrow = TyCon (TvName (TgInternal "(->)"))
```
SquidDev: Oh goodness.

src/Core/Optimise/Reduce.hs
zardyh: Good question.

src/Types/Infer/Builtin.hs
zardyh: I didn't actually use this :D:

src/Syntax/Resolve/Toplevel.hs
zardyh: Sure, I guess.

zardyh: Can you find the other instances of stuff like this?

src/Syntax.hs
```diff
@@ -265,9 +289,12 @@ makePrisms ''Lit
 makeLenses ''Skolem
 makeLenses ''TyBinder
 
+instance Spanned (Ann p) => Spanned (Binding p) where
+  annotation (Binding _ _ _ x) = annotation x
```
SquidDev: Could possibly be replaced with accessor.

src/Syntax/Let.hs
```diff
@@ -2,25 +2,26 @@
...
 
 freeIn :: (Show (Var p), Show (Ann p), Ord (Var p)) => Expr p -> Set.Set (Var p)
 freeIn (Ascription e _ _)   = freeIn e
 freeIn (RecordExt e rs _)   = freeIn e <> foldMap (freeIn . snd) rs
 freeIn (BinOp a b c _)      = freeIn a <> freeIn b <> freeIn c
 freeIn (VarRef v _)         = Set.singleton v
 freeIn (Begin es _)         = foldMap freeIn es
-freeIn (Let vs b _)         = (freeIn b <> foldMap (freeIn . snd3) vs) Set.\\ Set.fromList (map fst3 vs)
+freeIn (Let vs b _)         = (freeIn b <> foldMap (freeIn . e) vs) Set.\\ Set.fromList (map v vs) where
+  v (Binding v _ _ _) = v
```
SquidDev: Nope!

src/Syntax/Resolve.hs
```diff
@@ -115,11 +115,12 @@ resolveModule :: MonadResolve m => [Toplevel Parsed] -> m [Toplevel Resolved]
 resolveModule [] = pure []
 
 resolveModule (LetStmt vs:rs) =  do
-  let vars = map fst3 vs
+  let var (Binding v _ _ _) = v
```
SquidDev: Blughr

src/Syntax/Resolve.hs
```diff
@@ -198,13 +199,31 @@ lookupTy v = asks tyScope >>= lookupVar NotInScope v
...
+    (,) <$> resolveTele r as <*> reType r k
+  pure (TyVisArg v' k:as, (v, v'):vs)
+resolveTele _ [] = pure ([], [])
+
 reExpr :: MonadResolve m => Expr Parsed -> m (Expr Resolved)
 reExpr r@(VarRef v a) = flip VarRef a <$> (lookupEx v `catchJunk` r)
 
 reExpr (Let vs c a) = do
-  let vars = map fst3 vs
+  let var (Binding v _ _ _) = v
```
SquidDev: And here.

src/Syntax/Verify.hs
```diff
@@ -83,16 +82,17 @@ verifyProgram = traverse_ verifyStmt where
...
 verifyBindingGroup k _ = traverse_ verifyScc . depOrder where
-  verifyScc (AcyclicSCC (v, e, (s, t))) = do
+  verifyScc (AcyclicSCC (Binding v e _(s, t))) = do
     modify (k (BindingSite v s t))
     verifyExpr e
   verifyScc (CyclicSCC vs) = do
-    let vars = Set.fromList (map fst3 vs)
-    for_ vs $ \b@(var, ex, (s, ty)) -> do
+    let vars = Set.fromList (map var vs)
+        var (Binding v _ _ _) = v
```
SquidDev: Here.

src/Syntax/Verify.hs
```diff
@@ -146,7 +146,9 @@ unguardedVars (RecordExt e rs _)   = unguardedVars e <> foldMap (unguardedVars .
 unguardedVars (BinOp a b c _)      = unguardedVars a <> unguardedVars b <> unguardedVars c
 unguardedVars (VarRef v _)         = Set.singleton v
 unguardedVars (Begin es _)         = foldMap unguardedVars es
-unguardedVars (Let vs b _)         = (unguardedVars b <> foldMap (unguardedVars . snd3) vs) Set.\\ Set.fromList (map fst3 vs)
+unguardedVars (Let vs b _)         = (unguardedVars b <> foldMap (unguardedVars . e) vs) Set.\\ Set.fromList (map v vs) where
+  v (Binding v _ _ _) = v
```
SquidDev: Here.

zardyh: And.. fixed!

src/Syntax/Verify.hs
zardyh: Aa fixed!

src/Syntax/Resolve.hs
zardyh: Sorry, I don't know what I was thinking.

src/Syntax/Resolve.hs
zardyh: And this one..

src/Syntax/Let.hs
zardyh: Fixed this one too!

src/Syntax.hs
zardyh: Fixed!

zardyh:
@SquidDev I've updated the PR a fair bit - could you review again?

src/Types/Infer/Builtin.hs
```diff
@@ -156,6 +160,9 @@ litTy LiFloat{} = tyFloat
 -- A representation of an individual 'match' arm, for blaming type
 -- errors on:
 
+data SkipImplicit = DoSkip | Don'tSkip
```
SquidDev: Evil

PR #78, opened 2018-07-20
Unboxed tuples: Part 1

SquidDev:
It's almost midnight - I'm afraid I don't have energy to write a fancy commit message. Needless to say, the implementation is a little grim.

This implementation is intentionally very basic and restrictive:
 - Unboxed tuples can only (currently) appear on the left hand side of an arrow type. They cannot be returned.
 - We do not do any worker/wrapper transformation beyond binops within lower.
 - Codegen is probably broken.

This mainly intends to serve as a POC and a starting point for the codegen rewrite.

### DIFF
--- a/src/Backend/Lua/Emit.hs
+++ b/src/Backend/Lua/Emit.hs
@@ -94,11 +94,11 @@ import Syntax.Var
 import qualified Types.Wellformed as W
 
 
-type Returner = LuaExpr -> LuaStmt
+type Returner = [LuaExpr] -> LuaStmt
 
 data VarEntry v =
   VarEntry { vVar  :: v
-           , vExpr :: LuaExpr
+           , vExpr :: [LuaExpr]
            , vPure :: Bool
            }
   deriving (Show)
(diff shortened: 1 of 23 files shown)


src/Backend/Lua/Emit.hs
```diff
@@ -424,16 +428,24 @@ emitTerm (Match test branches)
 
   where withMatch p = map (\(a, b) -> VarEntry a b p)
 
+        isntTuple PatTupleUn{} = False
```
zardyh: You can have apostrophes in identifiers, y'know.

src/Core/Core.hs
```diff
@@ -42,15 +42,16 @@ data AnnTerm b a
   | AnnMatch b (AnnAtom b a) [AnnArm b a] -- ^ Pattern matching
 
   | AnnExtend b (AnnAtom b a) [(Text, Type a, AnnAtom b a)] -- ^ Record extension
+  | AnnTupleUn b [AnnAtom b a] -- ^ Unboxed tuple
```
zardyh: I don't like the name.

zardyh: Enable MagicHash everywhere and call it `AnnTuple#`?

zardyh: (That's not a serious suggestion but we really need a better name)

zardyh: (That's not a serious suggestion but we really need a better name)

SquidDev: I originally had `AnnUnboxedTup`, but it felt a wee bit too verbose. Dunno.

src/Backend/Lua/Emit.hs
zardyh: Then hlint is wrong!

src/Core/Core.hs
zardyh: Maybe `AnnValues`?

PR #79, opened 2018-07-28
Feature: Record Updates

zardyh:
I want to die.

This changes the typing rules (semantics?) of `{ x with r = e }` to be extension
of `x` to contain `r` but rather (monomorphic) update of `x`.

### DIFF
--- a/src/Syntax/Resolve.hs
+++ b/src/Syntax/Resolve.hs
@@ -34,6 +34,7 @@ import Control.Lens hiding (Lazy)
 
 import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
+import qualified Data.Set as Set
 import Data.Traversable
 import Data.Sequence (Seq)
 import Data.Foldable
@@ -61,6 +62,7 @@ data ResolveError
   | NoSuchModule (Var Parsed)
   | Ambiguous (Var Parsed) [Var Resolved]
   | NonLinearPattern (Var Resolved) [Pattern Resolved]
+  | NonLinearRecord (Expr Parsed) T.Text
   | EmptyMatch
   | EmptyBegin
 
@@ -74,10 +76,12 @@ instance Pretty ResolveError where
   pretty (NonLinearPattern v _) = "Non-linear pattern (multiple definitions of" <+> verbatim v <+> ")"
   pretty EmptyMatch = "Empty match expression"
   pretty EmptyBegin = "Empty begin expression"
+  pretty (NonLinearRecord _ t) = "Duplicate field" <+> stypeSkol (text t) <+> "in record" <#> empty
   pretty (ArisingFrom er ex) = pretty er <#> empty <#> nest 4 (string "Arising from use of" <+> blameOf ex </> pretty ex)
 
 instance Spanned ResolveError where
   annotation (ArisingFrom _ x) = annotation x
+  annotation (NonLinearRecord e _) = annotation e
   annotation _ = undefined
 
 instance Note ResolveError Style where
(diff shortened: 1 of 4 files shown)


zardyh:
The "enhancement" label is mildly disputable. But, hey, optics!

src/Syntax/Resolve.hs
```diff
@@ -293,11 +298,21 @@ reExpr r@(Ascription e t a) = Ascription
...
-reExpr (Record fs a) = Record <$> traverse (traverse reExpr) fs <*> pure a
-reExpr (RecordExt e fs a) = RecordExt
-                        <$> reExpr e
-                        <*> traverse (traverse reExpr) fs
-                        <*> pure a
+reExpr e@(Record fs a) = do
+  let ls = map fst fs
+      nub = Set.toList . Set.fromList
+  case ls \\ nub ls of
+    l:_ -> throwError $ NonLinearRecord e l
```
SquidDev: Can we do `tell (map (NonLinearRecord e) (nub (ll \\ nub ls)))` instead? This way we can report multiple errors.

src/Syntax/Resolve.hs
```diff
@@ -293,11 +298,21 @@ reExpr r@(Ascription e t a) = Ascription
...
+    _ -> pure ()
+  Record <$> traverse (traverse reExpr) fs <*> pure a
+reExpr (RecordExt e fs a) = do
+  let ls = map fst fs
+      nub = Set.toList . Set.fromList
+  case ls \\ nub ls of
+    l:_ -> throwError $ NonLinearRecord e l
```
SquidDev: Should this cover the whole expression, or is it OK just to include the extended part?

src/Syntax/Resolve.hs
SquidDev: Hrmr, is there a reason for defining a custom `nub`? I assume it's more efficient, but as `n` is probably tiny, I don't know how much difference it'd make.

There's a bit of me which would argue something like `mapMaybe (listToMaybe . tail) . group . sort`, though maybe that's a overboard.

src/Syntax/Resolve.hs
```diff
@@ -293,11 +298,21 @@ reExpr r@(Ascription e t a) = Ascription
...
                           <*> reType r t
                           <*> pure a
-reExpr (Record fs a) = Record <$> traverse (traverse reExpr) fs <*> pure a
-reExpr (RecordExt e fs a) = RecordExt
-                        <$> reExpr e
-                        <*> traverse (traverse reExpr) fs
-                        <*> pure a
+reExpr e@(Record fs a) = do
+  let ls = map fst fs
+      nub = Set.toList . Set.fromList
```
SquidDev: Hrmr, is there a reason for defining a custom `nub`? I assume it's more efficient, but as `n` is probably tiny, I don't know how much difference it'd make.

There's a bit of me which would argue something like `mapMaybe (listToMaybe . tail) . group . sort`, though maybe that's a overboard.

SquidDev: If you're just trying to find a list of duplicates, is that not ok?

src/Syntax/Resolve.hs
```diff
@@ -74,10 +76,12 @@ instance Pretty ResolveError where
   pretty (NonLinearPattern v _) = "Non-linear pattern (multiple definitions of" <+> verbatim v <+> ")"
   pretty EmptyMatch = "Empty match expression"
   pretty EmptyBegin = "Empty begin expression"
+  pretty (NonLinearRecord _ t) = "Duplicate field" <+> stypeSkol (text t) <+> "in record" <#> empty
```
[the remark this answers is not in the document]
zardyh: Yes.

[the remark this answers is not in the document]
zardyh: For some reason all resolve errors look silly and I can't quite grasp why

PR #81, opened 2018-08-06
Feature: Pattern Matching Lets

zardyh:
Now you can do

```
let (a, b) = ...
 in a + b
```

Fun! No more `match`es with a single branch! Yay!


zardyh:
You wrote that!

PR #82, opened 2018-08-07
Lex "trivial" tokens too

SquidDev:
This emits TcWhitespace and TcComment tokens when the "trivials" flag is set within the parser state. This hopefully allows for basic syntax highlighting just by emitting the stream of tokens.


zardyh:
it looked good

PR #83, opened 2018-08-09
Remove boolean operators

SquidDev:
These are not technically compiled correctly anyway (they are lazy when they should not be), so it would be better to handle this within the codegen.

That being said, the currently produced code is somewhat terrible: we need to eliminate lazy/force pairing and be smart about detecting such combinations of if/else blocks.

Note that this still keeps the operator precedence rules for `&&` and `||`.


zardyh:
For the title, surprisingly sane.

PR #84, opened 2018-08-09
Remove main

SquidDev:
We never did anything interesting with main, and our calling of it was rather broken, so it's probably better to remove it. People can now write this instead:

```ml
let () = f x
```

This is closer to how other MLs handle it, which is always nice. We can also remove some pretty ugly code from the backend.

The main disadvantage here is that it breaks a large number of our codegen tests. We used to use `main` to inject opaque values. As this now has no special meaning, it'll be stripped and so the test won't emit anything!

For now, I've fixed this by calling `main` with a `bottom : 'a` value:

```ml
let main f = ...

external val bottom : 'a = "nil"
let () = bottom (main bottom)
```

This definitely is pretty ugly (and can generate some awful code), but I'm OK with this for now. Hopefully the backend rewrite (which is increasingly reaching the status of Urn's resolver rewrite) will clean up a lot of these.

### DIFF
--- a/compiler/Main.hs
+++ b/compiler/Main.hs
@@ -64,7 +64,7 @@ compile opt (file:files) = runIdentity . flip evalNameyT firstName $ do
       optm <- case opt of
                 Do -> optimise lower
                 Don't -> pure (deadCodePass (reducePass lower))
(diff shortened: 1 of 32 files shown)


zardyh:
Looks good to me™

PR #89, opened 2018-08-22
Add a CSE pass

zardyh:
Pretty simple: We have a hashmap connecting expressions to their binding
sites, which we can use to increase sharing of pure variables.

For instance:

```ocaml
let x = Foo 1 in
let y = Foo 1 in
use x y
```

since x and y's values end up hashing to the same thing, we can
substitute all uses of y (since it's considered later) with a reference
to x, and generate

```ocaml
let x = Foo 1 in
use x x
```

### DIFF
--- a/src/Core/Optimise/CommonExpElim.hs
+++ b/src/Core/Optimise/CommonExpElim.hs
@@ -54,5 +54,7 @@ cseTerm scope map' (Let (Many vs) body) =
 
 worthIt :: Term a -> Bool
 worthIt Lam{} = False
+worthIt Let{} = False
 worthIt Atom{} = False
+worthIt Match{} = False
 worthIt _ = True
--- a/src/Core/Optimise/CommonExpElim.hs
+++ b/src/Core/Optimise/CommonExpElim.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE ScopedTypeVariables, ViewPatterns #-}
+-- | Eliminate common subexpressions by keeping a hashmap of terms to
+-- binding sites and recursively deduplicating each one-valued let.
+module Core.Optimise.CommonExpElim (csePass) where
+
+import qualified Data.HashMap.Strict as Map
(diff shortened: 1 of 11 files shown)


src/Data/VarMap.hs
```diff
@@ -9,8 +9,7 @@ module Data.VarMap
   , (<>), mempty
   ) where
 
-import qualified Data.IntMap.Strict as Map
-import qualified Data.IntMap.Merge.Strict as Map
+import qualified Data.HashMap.Strict as Map
```
SquidDev: Have we profiled this? I'm curious as to how the two perform.

zardyh: I haven't, but HashMaps are asymptotically more efficient than even PATRICIA trees. Since I was adding them anyway, I thought "why not change everything?".

src/Core/Optimise/CommonExpElim.hs
```diff
@@ -0,0 +1,58 @@
...
+cseTerm scope map' (Match ex as) = Match ex (map cseArm as) where
+  cseArm a = a & armBody %~ cseTerm scope map'
+cseTerm scope map' (Let (Many vs) body) =
+  let vs' = map (_3 %~ cseTerm scope map') vs
+      scope' = extendPureLets scope vs'
+   in Let (Many vs') (cseTerm scope' map' body)
+
+worthIt :: Term a -> Bool
+worthIt Lam{} = False
+worthIt Atom{} = False
```
SquidDev: Might be worth adding `Match` and `Let` here too?

zardyh: Sharing `match` would let us share record accesses, and `let` will be caught by commuting conversion in the reduce pass that immediately follows CSE, so I'm not bothered.

zardyh: (I did this, but GH is too stupid to notice)

PR #92, opened 2018-08-30
Lua quasi-quoter

SquidDev:
This adds a couple of things:

 - A parser for a subset of Lua. Our `Language.Lua.Syntax` data structures are missing a couple of things (namely `not`, `-`, `local function` and `foo:bar()`) so we do not yet parse those.
 - A Template Haskell quasi-quoter for Lua expressions and statements. One can use `$foo` to substitute variables, `%foo` for expressions and `@foo` for statements.
 - Replaces some of the more complex expressions in `Backend.Lua.Postprocess` with their quasi-quoted alternatives.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,9 @@
 
 src/Parser/Lexer.hs
 src/Parser.hs
-src/*.info
+src/Language/Lua/Parser/Lexer.hs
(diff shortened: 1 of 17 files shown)


zardyh:
lgtm :+1:

PR #97, opened 2018-09-11
Unboxing/uncurrying optimisation

SquidDev:
This doesn't do any of the fancy things (uncurrying function arguments, unpacking record arguments) yet, but it's a nice start.

### DIFF
--- a/tests/lua/stream.lua
+++ b/tests/lua/stream.lua
@@ -7,8 +7,7 @@ do
   local function Yield(x) return { __tag = "Yield", x } end
   local Done = { __tag = "Done" }
   local function Stream(x) return { __tag = "Stream", x } end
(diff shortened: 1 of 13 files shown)


tests/lua/stream.lua
```diff
@@ -7,16 +7,26 @@ do
...
   local go
   go = function(st)
+    local function yh(nk)
+      if nk.__tag == "Yield" then
+        local nu = nk[1]
+        io_write("'" .. to_string(nu._1) .. "', ")
+        return go(nu._2)
+      elseif nk.__tag == "Done" then
+        return print("]")
+      elseif nk.__tag == "Skip" then
```
zardyh: @SquidDev why?

PR #99, opened 2018-09-17
Use MonadChronicle for Infer

zardyh:
[muffled screaming]

### DIFF
--- a/compiler/Main.hs
+++ b/compiler/Main.hs
@@ -12,6 +12,7 @@ import qualified Data.Text as T
 import Data.Position (SourceName)
 import Data.Functor.Identity
 import Data.Foldable
+import Data.These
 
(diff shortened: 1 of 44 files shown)


compiler/Main.hs
```diff
@@ -79,21 +81,37 @@ compile opt (file:files) = runIdentity . flip evalNameyT firstName $ do
               desugared <- desugarProgram resolved
               infered <- inferProgram env desugared
               case infered of
-                Right (prog, env') ->
+                That (prog, env') ->
```
SquidDev: [OK](https://i.kym-cdn.com/photos/images/newsfeed/001/053/259/ebd.jpg)

It might be better to shift this into a `([TypeError], Maybe a)` or something instead - I'll have a look at that another time.

Issue #98, opened 2018-09-16
η-reduction in recursive binding groups not permissible


zardyh:
```ocaml
let () =
  let f x = f x
  f ()
```

Pretty simple code; An infinite loop. However, η-reduce `f` and you end up with

```lua
local f
f = f
f(__builtin_unit)
```

which is just `attempt to call a nil value (local 'f')`.

PR #100, opened 2018-09-22
An initial pass at cleaning up the reducer a little bit

SquidDev:
- Shift some code into separate modules, so it's hopefully a little easier to maintain.
 - Run the reducer inside a monad, and pass the scope around like that instead.
 - Keep track of when changes are made, and run the resolver multiple times until a fixed point is reached. 
 - Merge the `Joinify` into the reduce pass
 - Terms now handle traversal and reduction in the same pass, meaning we can eliminate some redundant visits (if, for instance, we know some term will be eliminated).


src/Core/Optimise/Reduce.hs
```diff
@@ -1,426 +1,273 @@
...
+
+            arms' <- traverse shoveJoinArm arms
+            changed $ Let (One (join, joinTy, Lam (TermArgument v ty) rest')) (Match test arms')
+
+        | otherwise -> pure (Let (One (v, ty, e')) rest')
+
+reduceTermK (Let (Many vs) rest) cont = do
+  -- We really can't do a lot with recursive lets sadly, so let's just
+  -- leave them for now.
+  vs' <- traverse (third3A reduceTerm) vs
```
zardyh: Will this η-reduce?

SquidDev: #98 is still present - that will be fixed when loop breakers are added.

PR #104, opened 2018-10-02
Some more inliner love

SquidDev:
Just a couple of improvements I'd like to make to the inliner and various optimisations

 - [x] Eliminate matches before determining if something is suitable for inlining.
 - [ ] Improve node costs for `scoreTerm`/`scoreAtom`, and discounts for applications and whatnot, so we don't inline large but trivial functions, and we do inline small but "worth it" ones.
 - [ ] Allow the uncurrying pass to either move matches on arguments into the a child lambda (problematic as it duplicates work) or unbox them directly.

The last of these is needed as `zip` gets translated into:
```ml
let zip = fun x ->
  match x with
  | Stream (f, start) -> fun y -> match y with
                       | Stream (g, start') -> (* The interesting stuff *)
```
Even after optimisation, we still have a match between the two lambdas, so we need to avoid that some how. It's worth noting that shifting the `match`es` further in does allow the unboxing to occur, but is not sufficient to get it to fully inline.

### DIFF
--- a/src/Core/Optimise/Reduce.hs
+++ b/src/Core/Optimise/Reduce.hs
@@ -434,7 +434,7 @@ reduceTermK _ (AnnMatch _ test arms) cont = do
   -- TODO: Work out a better way of handling continuations on multi-match
   -- arms, as our current handling within the let case means we have to
   -- walk down the entire tree.
-  case simplifyArms s test' arms of
+  case simplifyArms (underlying<$>) s test' arms of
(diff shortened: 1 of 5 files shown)


SquidDev:
Cherry picked the existing commit in d9b696570bbe13c17ffef58ec405de5efc5466cc. It'd be nice to work on this again in the future, but closing for now.

PR #105, opened 2018-10-04
REPL file loading

SquidDev:
- [x] Allow using `:l` (and `:load`) to load a file into the REPL.
 - [x] Add `:r` to reload already loaded files.

I'd like to add support for reloading files before finishing this off, but there's a coupel of questions worth resolving:

 - Should using `:load` reset the environment. This is useful, as there won't be artifacts from previous `:load`s.
 - If we do reset the environment, how does one load multiple files? One possibility would be to specify multiple files to `:load`. Obviously the whole module resolution thing is a whole 'nother discussion, but...
 - How should `:reload` handle file order (if we don't have variadic load)?

### DIFF
--- a/amuletml.cabal
+++ b/amuletml.cabal
@@ -23,6 +23,7 @@ executable amc
                      , amulet
                      , haskeline >= 0.7 && < 0.8
                      , containers >= 0.5 && < 0.6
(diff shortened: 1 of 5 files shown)


SquidDev:
I really need to clean up the REPL code at some point. Well, anything which needs to convert "plain text" into core.

PR #108, opened 2018-10-13
Feature: Type classes

zardyh:
yaaay


SquidDev:
Some things possibly worth handling now/in a later PR:
 - Parsing empty classes/instances
 - Termination checking in `resolve`
 - Nicer `MalformedInstance` error.
 - Checking all methods are defined in the class - I can't see an existing check for that.

zardyh:
> Checking all methods are defined in the class - I can't see an existing check for that.

I do believe it's called Core Lint

Issue #128, opened 2018-10-30
Pattern totallity checking


SquidDev:
We can probably implement [this paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/08/gadtpm-acm.pdf) to add a totality checker to the verify pass. It may also be worth utilising something similar in the optimiser, so we can strip GADT cases which we know will not happen.

PR #134, opened 2018-11-14
Pattern totality checking

SquidDev:
See #128

## Approach
Implements the a modified version of the algorithm described in [GADTs meet their match](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/08/gadtpm-acm.pdf). The key differences are:
 - Support for literals (infinite sets of possible constructors).
 - Compute the "covered" and "uncovered" set at the same time

## Summary of changes
 - Keep a mapping of type -> constructors within the `Env`ironment.
 - Move verification errors into a separate module.
 - Pass the `Env`ironment and a fresh variable into the verifier.


SquidDev:
It would be good to optimise the design of `Covering`, so we're not applying functions twice to pure elements. Maybe a `Seq (a, Kind)`, where `data Kind = Covered | Uncovered | Both`.

PR #139, opened 2018-12-19
Basic access modifiers

SquidDev:
It's simple, it's ugly, it needs tests.

 - [x] Private variables
 - [x] Private types
 - [x] Private modules.
 - [x] Private classes.
 - ~~Private class methods. To discuss: what are the semantics of this? I assume no-one will be able to implement classes with non-exported, non-default members?~~ Classes and module have a whole bundles of issues right now, so I'm leaving this off for now.
 - [x] Warn when private definitions are unused.


zardyh:
What about `private class`, `private instance`, `private module`, and this:
```
class foo 'a begin
  private val evil_method : forall 'b. 'a -> 'b
```
?

PR #140, opened 2018-12-20
List Comprehensions

zardyh:
eh


SquidDev:
One thing worth noting is that `[() | with a <- [1, 2], b <- sideeffect]` will cause `sideffect` to be evaluated twice. This definitely makes sense if you think about this as an alternative to `do` notation, but may still be considered surprising in some cases.

PR #142, opened 2018-12-20
Rewrite the test reporter to be less verbose and more sexy

SquidDev:
As the commit message says. Mostly inspired by every other test runner I've ever written:

 - Displays dots as the tests progress.
 - By default only shows failing tests/groups, though there are flags to make it more verbose (`--display t`, `--display g`).

Failures look like this:
image

Successes look like this:
image


zardyh:
This is great.

PR #152, opened 2019-06-11
Add monad syntax

zardyh:


SquidDev:
I'm still feeling a little guilty about tramping over `begin` in this context, but it doesn't serve much other purpose given the context system so not too worried.

👍

PR #153, opened 2019-06-13
CI Fun

SquidDev:
Jenkins is dead, so let's try to to get Azure Pipelines working instead.

It's me setting up a CI system, this may take some time...

There's no caching set up right now, so builds are painfully slow (20 minutes :/), but it's better than it was.


SquidDev:
Ouch - that's a half hour build time now, 90% of which is fetching/installing dependencies. Not sure if we'll stick with this.

PR #165, opened 2019-09-03
Add type synonyms

zardyh:
These are declared like

```ocaml
  type foo [args] <- expansion
```

and They Behave Like You Expect[citation needed]

### DIFF
--- a/src/Parser.y
+++ b/src/Parser.y
@@ -180,6 +180,11 @@ Top :: { Toplevel Parsed }
     | Access type BindName ListE(TyConArg) TypeBody { withPos2 $2 $3 $ TypeDecl $1 (getL $3) $4 $5 }
     | Access type TyConArg BindOp TyConArg TypeBody { withPos2 $2 $4 $ TypeDecl $1 (getL $4) [$3, $5] $6 }
 
+    | Access type BindName ListE(TyConArg) '<-' Type
+        { withPos2 $2 $3 $ TySymDecl $1 (getL $3) $4 (getL $6) }
+    | Access type TyConArg BindOp TyConArg '<-' Type
+        { withPos2 $2 $4 $ TySymDecl $1 (getL $4) [$3, $5] (getL $7) }
+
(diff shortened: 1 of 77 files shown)


src/Syntax/Toplevel.hs
```diff
@@ -30,6 +30,7 @@ data Toplevel p
   = LetStmt TopAccess [Binding p]
   | ForeignVal TopAccess (Var p) Text (Type p) (Ann p)
   | TypeDecl TopAccess (Var p) [TyConArg p] (Maybe [Constructor p]) (Ann p)
+  | TySymDecl TopAccess (Var p) [TyConArg p] (Type p) (Ann p)
```
SquidDev: Would it be better to have `TypeDecl` take something like `Empty | Alias (Type p) | Variant [Constructor p]`? Dunno if that'll actually simplify things or not.

src/Parser.y
```diff
@@ -180,6 +180,11 @@ Top :: { Toplevel Parsed }
     | Access type BindName ListE(TyConArg) TypeBody { withPos2 $2 $3 $ TypeDecl $1 (getL $3) $4 $5 }
     | Access type TyConArg BindOp TyConArg TypeBody { withPos2 $2 $4 $ TypeDecl $1 (getL $4) [$3, $5] $6 }
 
+    | Access type BindName ListE(TyConArg) '<-' Type
```
SquidDev: I assume the `<-` is to avoid conflicts with `type t = Foo`? I don't know if it's possible to have `type t = Foo` be a constructor, and `type t = (Foo)` be an alias, or if that's just going to be more confusing...

src/Syntax/Toplevel.hs
zardyh: I don't think this will simplify things because of how different kinding is for type declarations (always returns `*`, constructors need to have a kind ending in `*`, etc) vs type aliases (any return kind)

src/Parser.y
zardyh: This is a parser decision so I decided to leave it up to you.

PR #166, opened 2019-09-08
Feature: Functional dependencies

zardyh:
See the paper [Type Classes with Functional Dependencies](https://web.cecs.pdx.edu/~mpj/pubs/fundeps.html): the rules are implemented directly as in the paper.

This lets us specify the behaviour of type classes directly. For instance:
```ocaml
class r 'a 'b
class s 'a 'b | 'a -> 'b
class t 'a 'b | 'a -> 'b, 'b -> 'a
```

The class `r` acts like a (partial) relation, `s` is a partial function, and `t` is a partial 1-to-1 mapping. Concretely:

```ocaml
instance r int int
instance r string int
instance r int string
instance r string string
(* all ok *)

instance s int string
instance s string string
(* both ok *)
instance s int int
(* not ok - conflicts with s int string *)

instance t int int
instance t string string
(* both ok *)
instance t () int
instance t string ()
(* none ok *)
```


src/Types/Infer/Class.hs
```diff
@@ -433,20 +445,20 @@ reduceClassContext extra annot cons = do
...
       dedup scope ((var, con, r):needs)
-        | [ImplChoice _ t [] v _ _] <- lookup con scope =
+        | [ImplChoice _ t [] v _ _ _] <- lookup con scope =
           let (bindings, needs', scope') = dedup scope needs
            in if var == v
              then (bindings, needs', scope')
              else (Binding var (VarRef v (annot, t)) False (annot, t):bindings, needs', scope')
         | otherwise =
-          let (bindings, needs', scope') = dedup (insert annot LocalAssum var con scope) needs
+          let (bindings, needs', scope') = dedup (insert annot LocalAssum var con undefined scope) needs
```
SquidDev: Might be worth commenting why this can be `undefined`?

zardyh: aye

zardyh: donezoe

src/Syntax/Implicits.hs
```diff
@@ -45,7 +45,7 @@ deriving instance Eq (Var p) => Eq (Obligation p)
 
 -- | A concrete representation of /one/ implicit parameter, stored in an
 -- implicit parameter scope (a trie, indexed by the 'head').
-data Implicit p
+data Implicit info p
```
SquidDev: This always seems to be `ImplicitScope` - can we remove it at all, or is it just to break a dependency cycle?

zardyh: It's open just to break the cycle, yeah

PR #167, opened 2019-09-21
Feature: Type level literals

zardyh:
**Note**: Only `string` and `int` kinds for now, enforced badly by the parser. Technically nothing's stopping us from having the other literals in the type level, except for the fact that they don't have associated classes.

This expands upon the previous work on promoting things by promoting the `int` and `string` types to also be kinds. The inhabitants of those kinds being, of course, uninhabited. It also adds our first named built-in module:

```
module Amc = begin
  class known_int (x : int) begin
    val int_value : int
  end
  class known_string (x : int) begin
    val int_value : string
  end
end
```

As a demonstration for how expressive this feature is, I've included a very small Servant-inspired example of a type-safe HTTP router thing.


zardyh:
@SquidDev could you take care of the hlints?

src/Backend/Lua/Builtin.hs
```diff
@@ -118,6 +118,30 @@ builtins =
           return var[1]
         end
       |] )
+  , ( vStrVal, "__builtin_strval", [], Nothing
```
SquidDev: Ew, ew, ew.

src/Syntax/Implicits.hs
```diff
@@ -88,8 +90,28 @@ deriving instance (Show info, Show (Var p), Show (Ann p)) => Show (ImplicitScope
...
+  apply m (ManyMore is t) = ManyMore (apply m is) (apply m t)
+
+instance Ord (Var p) => Substitutable p (Implicit i p) where
+  ftv i = ftv (i ^. implType) Set.\\ boundByImpl i
+  apply m i = i & implHead %~ apply m' where
+    m' = m `Map.withoutKeys` boundByImpl i
+
 -- | Insert a choice for a *fully-known* (@Solved@) implicit parameter
--- (the variable @v@) of type @tau@ at the given trie.
+-- (the vriable @v@) of type @tau@ at the given trie.
```
SquidDev: Can we revert this?

tests/parser/fail_span.ml
```diff
@@ -1 +0,0 @@
-let main : "foo" = 1
```
SquidDev: Can we add this test back in, just with something else? `let main : match` will be good.

src/Syntax/Implicits.hs
zardyh: yes.

src/Backend/Lua/Builtin.hs
zardyh: You come up with a better way to do it, then :/

tests/parser/fail_span.ml
zardyh: sure

PR #168, opened 2019-09-22
WIP: Add a row_cons class

zardyh:


src/Backend/Lua/Builtin.hs
```diff
@@ -118,30 +118,62 @@ builtins =
...
       |] )
+  , ( vROWCONS, "__builtin_rowcons", [], Just (1, \[l] -> (mempty, [ [lua|%l|] ]))
+    , [luaStmts|
+        local function __builtin_rowcons(x)
+          return x
+        end
+      |] )
+  , ( vExtend, "__builtin_extend", [], Nothing
+    , [luaStmts|
+        local function __builtin_extend(key, value, record)
```
SquidDev: I wonder if it's worth having a generic `__builtin_copy` function, which both extend and restrict depend on? I could also get the codegen to emit a call to it, instead of inlining the loop.

src/Backend/Lua/Builtin.hs
```diff
@@ -118,30 +118,62 @@ builtins =
...
+          local new = {}
+          for k, v in pairs(record) do
+            new[k] = v
+          end
+          new[key] = value
+          return new
+        end
+      |] )
+  , ( vRestrict, "__builtin_restrict", [], Just (2, \[key, t] -> (mempty, [[lua| { _1 = %t[%key], _2 = %t }|]]))
```
SquidDev: Ok, this inlined version is super dubious, but I agree with your reasoning. Though I dunno if it'll cause issues with Lua interop in the future.

zardyh: That's true, the FFI could observe our terrible optimisation, but I claim that's Undefined Behaviour.

PR #169, opened 2019-09-24
[WIP] Feature: Term Inference

zardyh:
The compiler is smart enough to suggest fits for some holes. In fact, it's smart enough to prove [several sentences of first order logic](https://github.com/tmpim/amulet/blob/feature/infer-terms/tests/types/holes.ml).

The search algorithm is really simple, too! It's great. Please :eyes: this PR carefully though.


SquidDev:
As somewhat mentioned on Slack, there's a couple of cases which are a little broken:

### Literals, when no appropriate variables are in scope
```
=stdin[1:2 ..1:2]: error
  │
1 │ (_ : int)
  │  ^
  Found typed hole _ of type int
  • Valid replacements include:
      • int_value
      • ** int_value int_value
      • ** int_value (** int_value int_value)
```

However, `int_value` isn't in scope, so this provided code fails to work (also not entirely clear why this only infers the `**` operator - I assume this is due to which symbol it tries first?).

### Record extensions
```
  │
1 │ let f : forall 'a 'b. { 'b | x : int } -> int -> { 'b | x : int } = _ 
  │                                                                     ^ 
  Found typed hole _ of type { fi'b | x : int } -> int -> { fi'b | x : int } 
  • Valid replacements include:
      • fun { x = y } -> fun z -> { x = z }
      • fun { x = y } -> fun z -> { x = int_value } 
      • fun { x = y } -> fun z -> { x = ** z z }
```

All of these have the type `{ 'a | x : 'b } -> 'c -> { x : 'c }` instead, which is slightly different but wrong. I'm perfectly fine with just removing the record extension case for now - I'd rather avoid generating incorrect code.

### Holes not getting the correct type?

This works as expected:
```
  │       
1 │  fun (x : 'a) -> (_ : 'a)
  │                   ^    
  Found typed hole _ of type 'a
  • Valid replacements include:
      • x
```

However, adding an additional argument does not:
```
  │
1 │  fun (x : 'a) _ -> (_ : 'a) (* Alternatively fun _ (x : 'a) *).
  │                     ^
  Found typed hole _ of type 'id
```
I'm afraid I'm not quite sure why this happens - the ascription TC code all looks reasonable to me, but I still don't quite understand the intricacies.

PR #171, opened 2019-09-26
Dictionaries in GADTs + Refinements to the solver

zardyh:
Internally, `amc` now supports constructors with vectors of arguments.
The surface language only exposes this by letting us put dictionaries
inside GADTs, like God intended:

```ocaml
type dict 'c = Dict : 'c => dict 'c
type 'p :- 'q = Sub of ('p => unit -> dict 'q)
(* Entailment - morphisms in the category Con *)
```

`dict` is a newtype with the interesting property that `dict 'c ~ c`;
Matching on it will reveal the evidence for `'c` that was hidden away.

This lets us talk about the category of constraints, `Con`, inside the
language, which revealed a treasure trove of weird edge cases in the
constraint solver which I have spent the last of my sanity fixing.

For instance, `Con` is cartesian: You can ignore information. This
didn't work until 5144968, but it does now, so all good.

```ocaml
let weaken1 : forall 'a 'b. ('a * 'b) :- 'a = Sub (fun _ -> Dict)
let weaken2 : forall 'a 'b. ('a * 'b) :- 'b = Sub (fun _ -> Dict)
```


src/Syntax/Expr.hs
```diff
@@ -234,7 +233,7 @@ isFn _ = False
 conVarRef :: Var a ~ Var Resolved => Expr a -> Bool
 conVarRef (VarRef t _) =
   case t of
-    TgName t _ -> isUpper (T.head t)
+    TgName t i -> i == (-36) || isUpper (T.head t) -- -36 is lAZYName
```
SquidDev: We really need a better way of doing this, especially as constants are less fixed now.

zardyh:
You're right, it is impossible.

src/Syntax/Expr.hs
zardyh: Thanks :)

PR #173, opened 2019-09-27
Add a REPL client/server

zardyh:
Amc REPLs will listen (on port 5478, pass port 0 to disable), and clients can connect and run code there a lot faster than spinning up a new REPL[citation needed].


compiler/Repl.hs
```diff
@@ -452,19 +468,96 @@ execFile name line = do
...
+
+      handleReplLine line =
+        if | T.null line -> pure ()
+           | ':' == T.head line -> uncurry (execCommand Nothing) . span (/=' ') $ T.unpack (T.tail line)
+           | otherwise -> () <$ execString "=<remote command>" line
+
+
+  nulldev <- openFile "/dev/null" WriteMode
+  bracket getSock Net.close $ \sock -> do
+    state <- execStateT (loadFiles files) (state { outputHandle = nulldev })
```
SquidDev: OK, this seems crazy dubious.

zardyh: oh yeah it really is but I couldn't think of any better way. we don't want it to print the loading message twice (once for the server, once for the repl itself), so I just did that instead

zardyh: ah quick fix load once before starting either I suppose

zardyh: gimme a second I'm on it

PR #174, opened 2019-09-29
Quick POC for suggesting you convert matches to functions.

SquidDev:
There's some pretty gross stuff going on here due to `function` being desugared away. But it works, I guess??

### DIFF
--- a/src/Syntax/Verify.hs
+++ b/src/Syntax/Verify.hs
@@ -114,16 +114,23 @@ verifyExpr ex@(Let vs e (_, ty)) = do
   verifyExpr e
 verifyExpr (If c t e _) = traverse_ verifyExpr [c, t, e]
 verifyExpr (App f x _) = verifyExpr f *> verifyExpr x
-verifyExpr m@(Fun (PatParam p) x _) = verifyMatch m (getType p) [Arm p Nothing x]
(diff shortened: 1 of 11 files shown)


zardyh:
Nice! Is this mergeable?

src/Syntax/Verify.hs
```diff
@@ -114,16 +114,23 @@ verifyExpr ex@(Let vs e (_, ty)) = do
   verifyExpr e
 verifyExpr (If c t e _) = traverse_ verifyExpr [c, t, e]
 verifyExpr (App f x _) = verifyExpr f *> verifyExpr x
-verifyExpr m@(Fun (PatParam p) x _) = verifyMatch m (getType p) [Arm p Nothing x]
+verifyExpr m@(Fun (PatParam p) x _) = verifyMatch (const $ pure ()) m (getType p) [Arm p Nothing x]
+verifyExpr m@(Fun (EvParam _) (Match e bs _) _) = do
```
zardyh: oof

SquidDev: Yeah. I don't think we'll ever inject `ExprWrapper` here or anything, so it _should_ be fine? Maybe...

PR #176, opened 2019-10-04
Feature: Type Functions

zardyh:
Support for closed type functions with overlapping equations based on
the paper _Closed Type Families with Overlapping Equations_ by Richard
Eisenberg &co.

Type functions declared like this:

```ocaml
type function foo 'a begin
  ...
end
```

where the `...` can contain any number of clauses of the form

```ocaml
 foo 'a = ... 'a ...
```

Type function clauses can be _overlapping_:

```ocaml
type function equal 'a 'b : bool_t begin
  equal 'a 'a = True
  equal 'a 'b = False
end
```

But that's ok, we still have soundness:

```
soundness.ml[20:76 ..20:79]: error
  No instance for equal 'a 'b ~ false arising from use of the expression
   │ 
20 │ let no : forall 'a 'b. trivial :- (equal 'a 'b ~ string) = Sub (fun _ -> Dict)
   │                                                                          ^^^^
```

We only take a branch when it's _apart_ from all the other branches.
`apart(p, t)` is a predicate defined by `¬unify(p, flatten(t))`, where
`flatten` is an operation that replaces all type family applications by
fresh type variables.

If two branches overlap and wouldn't be apart at matching time, we error
if their right-hand sides are incompatible:

```
overlap.ml[2:3 ..2:18]: error
  Overlapping equations for foo int
  • Note: first defined here,
  │ 
2 │   foo int = string
  │   ^^^^^^^^^^^^^^^^
   but also defined here
  │ 
3 │   foo int = bool
  │   ^^^^^^^^^^^^^^
```

There's a built-in type function called `type_error` that lets the user
write custom type errors using a type-level DSL that reflects the
compiler's `Doc` type:

```ocaml
instance type_error (String "Can't show functional type" :<>: ShowType ('a -> 'b))
      => show ('a -> 'b)
begin
  let show x = show x
end
```

```
tyfun-err02.ml[13:9 ..13:29]: error
   │ 
13 │ let _ = show (fun x -> x + 1)
   │         ^^^^^^^^^^^^^^^^^^^^^
  Can't show functional type int -> int
```


src/Parser/Context.hs
```diff
@@ -353,12 +355,27 @@ handleContextBlock needsSep  tok@(Token tk tp te) c =
...
               , ck)
 
+    -- @type function ... begin@ ~~> Replace type head with body context
+    (TcBegin, CtxTypeHead offside:ck)
+      | spCol tp >= spCol offside
+      -> pure (Result tok Done, CtxEmptyBlock Nothing : CtxTypeFunBody : CtxBracket TcEnd : ck)
+
+    (tc, CtxTypeHead cls:ck)
+      | spCol tp > spCol cls
+      , isTopTok tc && tc /= TcType
```
SquidDev: Why the `tc /= TcType`?

SquidDev: In fact, why this rule at all? It's not clear when it will be triggered given the syntax for type functions.

zardyh: idk i don't know this stuff lol it just didn't work without the tc /= TcType

PR #177, opened 2019-10-04
Feature: Associated Type Families

zardyh:
Conservative extension of #176 to support associated type families as part of type class declarations. Yay!


src/Control/Monad/Chronicles.hs
```diff
@@ -60,7 +62,7 @@ absolving l r = memento l >>= either (const r) pure
 
 -- | Remove any record that a computation had. If it ended with
 -- 'confess', explode.
-silence :: MonadChronicle c m => m a -> m a
+silence :: (HasCallStack, MonadChronicle c m) => m a -> m a
```
SquidDev: I don't know what this is, and I don't like it :p.

zardyh: Propagates the error up so `silence` will tell you where it was called

PR #181, opened 2019-10-11
Add idiom brackets

zardyh:
`(| f x y z |)` = `pure f <*> x <*> y <*> z`; Like `begin`-expressions, uses whatever `pure` and `<*>` are in scope.

### DIFF
--- a/src/Parser.y
+++ b/src/Parser.y
@@ -332,7 +332,7 @@ Atom :: { Expr Parsed }
      | hole                                   { withPos1 $1 (Hole (Name (getHole $1))) }
      | '_'                                    { withPos1 $1 (Hole (Name (T.singleton '_'))) }
      | begin List1(CompStmt, ExprSep) end     { withPos2 $1 $3 $ DoExpr bindVar $2 }
-     | '(|' ListE1(PreAtom) '|)'              { withPos2 $1 $3 $ Idiom pureVar apVar $2 }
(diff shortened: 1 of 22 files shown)


src/Parser.y
```diff
@@ -330,6 +332,7 @@ Atom :: { Expr Parsed }
      | hole                                   { withPos1 $1 (Hole (Name (getHole $1))) }
      | '_'                                    { withPos1 $1 (Hole (Name (T.singleton '_'))) }
      | begin List1(CompStmt, ExprSep) end     { withPos2 $1 $3 $ DoExpr bindVar $2 }
+     | '(|' ListE1(PreAtom) '|)'              { withPos2 $1 $3 $ Idiom pureVar apVar $2 }
```
SquidDev: For consistency, possibly should be `Expr0 ListE(PreAtom)` ?

src/Parser/Token.hs
```diff
@@ -151,6 +153,8 @@ instance Show TokenClass where
   show TcCBrace = "}"
   show TcOSquare = "["
   show TcCSquare = "]"
+  show TcOBanana = "(|"
+  show TcCBanana = "|)"
```
SquidDev: Can you update `compiler/Errors.hs`, so that the bounds in `highlightAmulet` are correct? Sorry, I realise it's ugly.

zardyh: Of course

src/Parser.y
zardyh: W-why?

zardyh: oh, for consistency with application? sure

PR #184, opened 2019-10-12
Guillotine the Built-ins

zardyh:
Move everything that the TC doesn't have to care about to a `lib/prelude.ml` file. This file is looked up in:

* AMC_PRELUDE
* AMC_LIBDIR/prelude.ml
* relative to the compiler

this was pain


zardyh:
@SquidDev help
```
> open import "%prelude"

> open import "%option"

> (+1) <$> Some 1
=stdin:33: attempt to call a nil value (global 'Functor')
```

zardyh:
nvm

src/Frontend/Driver.hs
```diff
@@ -390,6 +413,7 @@ getTypeEnv path = do
                     & (names %~ (<> (modEnv ^. names)))
                     . (types %~ (<> (modEnv ^. types)))
                     . (classDecs %~ (<> (modEnv ^. classDecs)))
+                    . (classes %~ (<> (modEnv ^. classes)))
```
SquidDev: Hrmr, does this mean class instances are exposed too? This code was originally copied from `Types.Infer`.

src/Frontend/Driver.hs
```diff
@@ -261,32 +268,48 @@ compile path = do
...
+          files = map (view _1 . get_vertex) vertices
+      pure (nub (reverse files))
 
 lowerIt :: (MonadNamey m, MonadState Driver m)
         => Bool -> FilePath
         -> (Set.Set FilePath, LowerState, Seq.Seq [Stmt CoVar])
         -> m (Set.Set FilePath, LowerState, Seq.Seq [Stmt CoVar])
 lowerIt all path (visited, lEnv, stmts)
-  | path `Set.member` visited = pure (visited, lEnv, stmts)
+  -- | path `Set.member` visited = pure (visited, lEnv, stmts)
```
SquidDev: Can we remove the visited flag entirely then? I think my original code was technically fine, it should have just been recursing and wasn't for some reason :/.

tests/types/multi_errs01.out
```diff
@@ -1,18 +1,6 @@
-multi_errs01.ml[2:19 ..2:22]: error
+multi_errs01.ml[3:19 ..3:22]: error
   │ 
-2 │   let foo x = 2 + true
+3 │   let foo x = 2 + true
   │                   ^^^^
   Couldn't match actual type bool
     with the type expected by the context, int
-multi_errs01.ml[3:3 ..3:11]: error
```
SquidDev: Should I be worried?

tests/verify/match_to.out
```diff
@@ -1,26 +1,37 @@
-match_to.ml[1:9 ..1:16]: note
+definition of +[1:1 ..1:1]: warning
```
SquidDev: Can we fix all the syntax errors within the verify tests at least? Even if it still generates bogus code like `nil` (though ideally it's valid).

tests/resolve/pass_ops.out
```diff
@@ -1,5 +1,14 @@
-let |>#0 x#1 f#2 = f#2 x#1
-let <|#3 f#4 x#5 = f#4 x#5
-let add#6 a#7 b#8 = (a#7 + b#8)
-let main#9 =
-  (((2 |>#0 ((+ 3)))) add#6 3)
+pass_ops.ml[4:19 ..4:19]: error
```
SquidDev: Can we fix these tests? A lot of the resolve ones are failing, when they really shouldn't be.

tests/lua/match_consumed.lua
```diff
@@ -3,7 +3,7 @@ do
     return function(x)
       local a = f(1)
       if x.__tag == "None" then
-        return f(a)
+        return f(a + 0)
```
SquidDev: Uggggghr. It's fine, we can discuss how to handle this later.

map-compose.ml
```diff
@@ -0,0 +1,58 @@
+type nil
```
SquidDev: I assume this shouldn't be here?

tests/lua/opt_unbox_args.ml
```diff
@@ -1,3 +1,7 @@
+external val ( * ) : int -> int -> int = "function(x, y) return x * y end"
+external val ( - ) : int -> int -> int = "function(x, y) return x * y end"
+external val ( <= ) : int -> int -> bool = "function(x, y) return x > y end"
```
SquidDev: 🤔

src/Frontend/Driver.hs
zardyh: Yes

map-compose.ml
zardyh: probably not

tests/lua/opt_unbox_args.ml
zardyh: :thinking:

tests/types/multi_errs01.out
zardyh: probably not

tests/verify/match_to.out
zardyh: done

tests/resolve/pass_ops.out
zardyh: done

tests/lua/match_consumed.lua
zardyh: won't do

tests/lua/opt_unbox_args.ml
zardyh: fixed

src/Frontend/Driver.hs
SquidDev: OK, this feels a little problematic, but I can live - I guess we can fix it when I sort out prelude loading for `compile`.

Issue #187, opened 2019-10-13
Error Handling


zardyh:
What do? Exceptions? Monads?

zardyh:
[Here](https://hydraz.semi.works/txt/exception-prototype.ml.html)'s a prototype of what Amulet exceptions might look like. The `typeable` class and its instances need to be compiler magic, but other than that everything is implementable in library code.

zardyh:
Amulet exceptions landed in aad9de410f2d72326d56b51c42c9dfacd2712375 and they were fixed to ignore Lua exceptions in 2a22dd4c5988e943fc590a04626567f50a315ed6, so this issue is fixed (and I forgot to close it)

PR #188, opened 2019-10-13
Bind to some Lua libraries

zardyh:
These are all the ones I could bind to without figuring out #187 first.


lib/lua/math.ml
```diff
@@ -0,0 +1,39 @@
...
+external val pi : float = "math.pi"
+
+external val positive_infty : float = "math.huge"
+external val negative_infty : float = "-math.huge"
+external val nan            : float = "-(0/0)"
+
+external val is_nan : float -> bool =
+  "function(x) return (type(x) == 'number') and x ~= x end"
+
+external val seed_random  : float -> unit = "math.randomseed"
```
SquidDev: [This should probably be](https://www.lua.org/source/5.3/lmathlib.c.html#math_randomseed) `int -> unit`.

lib/lua/os.ml
```diff
@@ -0,0 +1,30 @@
...
+     end\
+   end"
+
+external val exit_success : unit -> 'a =
+  "function() return os.exit(0) end"
+
+external val exit_failure : unit -> 'a =
+  "function() return os.exit(1) end"
+
+external val getenv : string -> string =
```
SquidDev: Hrrm, would it be worth this returning an `option string` instead? I realise this'd require some additional work, as we'd need to have a way of checking and coercing values.

If not, maybe just `os.getenv(x) or ""`? Should mean we can inline it too.

lib/prelude.ml
```diff
@@ -110,3 +110,106 @@ instance monad list begin
...
+  let xs >>= f =
+    match xs with
+    | Some x -> f x
+    | None -> None
+end
+
+external val parse_int : string -> option int =
+  "function(x) \
+     if tonumber(x) and ((math.modf(tonumber(x))) == tonumber(x)) then \
+       return Some(tonumber(x)) \
```
SquidDev: So strictly speaking, this may not be safe - `Some`/`None` may be renamed. I'm not 100% sure it'll even be emitted even - only if people actually construct values using it (though would need to check).

I think the "cleaner" thing to do here would be to have some `Obj.magic`-esque function and expose `type` - then we can have some function from `lua_repr -> option 'a` (clearly unsound, but hopefully used in a safe manner).

lib/lua/math.ml
```diff
@@ -0,0 +1,39 @@
...
+   end"
+
+external val pi : float = "math.pi"
+
+external val positive_infty : float = "math.huge"
+external val negative_infty : float = "-math.huge"
+external val nan            : float = "-(0/0)"
+
+external val is_nan : float -> bool =
+  "function(x) return (type(x) == 'number') and x ~= x end"
```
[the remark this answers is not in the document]
zardyh: True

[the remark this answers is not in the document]
zardyh: @SquidDev Also: The parser's broken, here, remove the parens and it parses as

```
type(x) == ('number' and x) ~= x
```

lib/lua/math.ml
zardyh: Well observed, I just assumed it took any number from looking at the docs!

lib/lua/os.ml
zardyh: ... Aye, I forgot about `or`. I don't return an `option string` here because the `lua/` modules are all "freestanding" - they don't depend on `prelude.ml`. The option is unboxing the `option` return value and taking two continuations

lib/prelude.ml
zardyh: I can make it `private` then pass the constructors to it, which is a better stopgap solution here if you ask me

SquidDev: Ohh, that's simple and clean. 👍

lib/lua/math.ml
zardyh: Done

PR #190, opened 2019-10-13
Lots of minor improvements to the module system

### LINKED ISSUE #189: Errors in loaded files get forgotten
... and this leads to a compiler crash.

### Setup
a.ml:
```
let () = 1
```
b.ml:
```
open import "a.ml"
```
Now compile `b.ml`:
```
*** Exception: Prelude.undefined
CallStack (from HasCallStack):
  error, called at libraries/base/GHC/Err.hs:80:14 in base:GHC.Err
  undefined, called at ./src/Control/Monad/Chronicles.hs:41:19 in main:Control.Monad.Chronicles
```

SquidDev:
I was going to do the prelude loading in the same changset, and then realised I should probably get on with some uni work. So here goes for now... 

---

 - Abandon our approach of only trying to show new errors - most of the
   time, it's actually more useful to provide every error available.

   This does mean that warnings will show up each time a user imports a
   file multiple times within the REPL, but I hope that's rare enough
   that this ends up being a satisfactory fix.

   Closes #189

 - Both the emitted core and errors are now gathered with a depth-first
   seach over the dependency graph. This hopefully simplifies things a
   little.

 - Don't crash when a module is missing, instead just return 'Nothing' -
   while we were correctly handling missing files within 'importModule',
   they'd still be added to the graph, which caused issues when
   gathering errors later on.

 - Print import loops more elegantly. I meant to do this when originally
   implementing it, and then entirely forgot.

### DIFF
--- a/src/Frontend/Driver.hs
+++ b/src/Frontend/Driver.hs
@@ -289,7 +289,7 @@ gatherDepsOf f = fmap snd . foldlM go mempty where
     | path `Set.member` visited = pure (visited, seq)
     | otherwise = do
         this <- f path
-        deps <- uses (modules . at path) (maybe mempty (^.dependencies)) -- mathues: I don't understand lens.
+        deps <- uses (modules . at path) (foldMap (^.dependencies))
         (visited, seq) <- foldlM go (Set.insert path visited, seq) deps
         pure (visited, seq Seq.|> this)
 
@@ -445,7 +445,7 @@ getVerifiedAll :: (MonadNamey m, MonadState Driver m, MonadIO m)
 getVerifiedAll path = do
   here <- getVerified path
 
-  deps <- uses (modules . at path) (foldMap (^.dependencies))
+  deps <- uses (modules . at path) (foldMap (^. dependencies))
   there <- foldMapM getVerifiedAll deps
 
   pure (All here <> there)
--- a/src/Frontend/Driver.hs
+++ b/src/Frontend/Driver.hs
@@ -40,7 +40,7 @@ module Frontend.Driver
   -- $query
   , getSignature, getTypeEnv
   , getVerified, getVerifiedAll
-  , getLowered
+  , getLowerState, getLowered
   , loadPrelude
   ) where
 
@@ -61,15 +61,12 @@ import qualified Data.Sequence as Seq
 import qualified Data.Text.IO as T
 import qualified Data.Text as T
 import qualified Data.Set as Set
-import Data.Traversable
-import Data.Bifunctor
 import Data.Position
 import Data.Foldable
 import Data.Functor
-import Data.Graph
+import Data.Monoid
 import Data.Maybe
 import Data.These
-import Data.List
 import Data.Span
 
 import Core.Core (Stmt)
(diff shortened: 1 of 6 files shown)


src/Frontend/Driver.hs
```diff
@@ -264,52 +270,34 @@ lowerWith root parsed sig env lState = do
...
+
+gatherDepsOf :: (MonadNamey m, MonadState Driver m)
+             => (FilePath -> m a)
+             -> Set.Set FilePath -> m (Seq.Seq a)
+gatherDepsOf f = fmap snd . foldlM go mempty where
+  go (visited, seq) path
+    | path `Set.member` visited = pure (visited, seq)
+    | otherwise = do
+        this <- f path
+        deps <- uses (modules . at path) (maybe mempty (^.dependencies)) -- mathues: I don't understand lens.
```
zardyh: `uses (modules . at path . dependencies . non mempty)`

src/Frontend/Driver.hs
```diff
@@ -355,49 +343,53 @@ addModule path = do
...
-    Nothing -> addModule path
-    Just mod -> pure mod
+    Nothing -> do
+      exists <- liftIO $ doesFileExist path
+      if exists then Just <$> addModule path else pure Nothing
+    Just mod -> pure (Just mod)
 
-  case mod ^. stage of
-    SUnparsed -> pure (Nothing, mod ^. errors)
+  case maybe SUnparsed (^. stage) mod of
```
zardyh: `mod ^. state . non SUnparsed`

src/Frontend/Driver.hs
```diff
@@ -355,49 +343,53 @@ addModule path = do
...
-  (_, depErrors) <- getSignature path
+  _ <- getSignature path
 
-  mod <- findMod path
-  case mod ^. stage of
-    SParsed{} -> pure (Nothing, depErrors)
-    SUntyped{} -> pure (Nothing, depErrors)
-    SUnresolved{} -> pure (Nothing, depErrors)
+  mod <- use (modules . at path)
+  case maybe SUnparsed (^.stage) mod of
```
zardyh: `mod ^. stage . non SUnparsed`

src/Frontend/Driver.hs
```diff
@@ -407,71 +399,89 @@ getTypeEnv path = do
...
-  (_, depErrors) <- getTypeEnv path
+  _ <- getTypeEnv path
 
-  mod <- findMod path
-  case mod ^. stage of
-    SParsed{} -> pure (Nothing, depErrors)
-    SUnparsed{} -> pure (Nothing, depErrors)
-    SUntyped{} -> pure (Nothing, depErrors)
-    SUnresolved{} -> pure (Nothing, depErrors)
+  stg <- uses (modules . at path) (maybe SUnparsed (^.stage))
```
zardyh: `use (modules . at path . stage . non SUnparsed)`

src/Frontend/Driver.hs
```diff
@@ -407,71 +399,89 @@ getTypeEnv path = do
...
 getVerifiedAll path = do
-  (here, errors) <- getVerified path
-
-  mod <- findMod path
-  (there, depErrors) <- checkAll getVerifiedAll () (mod ^. dependencies)
-
-  pure (here >> there, depErrors <> errors)
+  here <- getVerified path
+
+  deps <- uses (modules . at path) (foldMap (^.dependencies))
```
zardyh: `use (modules ^. at path . dependencies)` should be fine?

src/Frontend/Driver.hs
```diff
@@ -618,18 +629,14 @@ instance Semigroup a => Semigroup (AllOf a) where
...
-    = first (\(AllOf x) -> x)
-  <$> foldrM (\x acc -> (<>acc) . first AllOf <$> fn x)
-             (AllOf (Just start), mempty) xs
-
--- | Find a module within the state, knowing that it is present.
-findMod :: MonadState Driver m => FilePath -> m Module
-findMod p = uses modules (fromJust . Map.lookup p)
+foldMapM :: (Monad m, Foldable t, Monoid b)
+         => (a -> m b) -> t a -> m b
+foldMapM f = foldrM (\a b -> (<>b) <$> f a) mempty
```
zardyh: beautiful [sheds a single tear]

src/Syntax/Resolve/Error.hs
```diff
@@ -73,7 +73,7 @@ instance Pretty ResolveError where
   pretty LetOpenStruct = "Cannot declare a module within a" <+> keyword "let open" <+> "expression"
 
   pretty (UnresolvedImport name) = "Cannot resolve" <+> dquotes (text name)
-  pretty (ImportLoop ((name, _) E.:| _)) = "Loop loading " <+> dquotes (string name)
+  pretty (ImportLoop ((name, _) E.:| _)) = "Cycle importing" <+> dquotes (string name)
```
zardyh: What about something descriptive like "modules form an import cycle"?

PR #199, opened 2019-10-17
A bit of a reorganisation

SquidDev:
This is a small attempt to clean up where a couple of files are located. I'm not especially attached to any of these changes, so feel free to close/suggest something different.

 - Move all executables into the "bin" folder (merges "compiler" and "tools" directories).
 - Move test-related code into "tests/driver".


zardyh:
lgtm

PR #203, opened 2019-10-19
Quick Look Impredicativity

zardyh:
This lets us infer /some/ uses of impredicative types.


src/Types/Infer/App.hs
```diff
@@ -0,0 +1,224 @@
...
+import Types.Unify
+
+import Text.Pretty.Semantic
+
+
+#ifdef TRACE_TC
+import Debug.Trace
+#endif
+
+-- Infer an application, supporting "Quick Look" impredicativity: We
```
SquidDev: Possibly make this `-- |`, so it's a doc comment?

src/Types/Infer/App.hs
```diff
@@ -0,0 +1,224 @@
...
+    r_ql_sub <- case expected of
+      Just tau@(TyApps (TyCon _) (_:_)) | result@(TyApps (TyCon _) (_:_)) <- getQuantR quantifiers -> do
+-- Pushing down the result type into the quick look substitution
+-- is only sound if both are /guarded/. Do note that missing a
+-- substitution here isn't unsound, it'll just lead to wonky inference.
+        _ <- subsumes (becauseExp exp) result (apply ql_sub tau)
+        pure $ fromMaybe mempty (unifyPure result tau)
+      _ -> pure mempty
+
+#ifdef TRACE_TC
```
SquidDev: Is it worth refactoring the trace code from `Types.Unify` into a `Types.Trace` module? Then we can share it everywhere, rather than using the CPP in some places, and not in others.

src/Types/Unify.hs
```diff
@@ -460,15 +461,17 @@ bind scope var ty
   | TyVar var == ty = pure (ReflCo ty)
   -- /\ Var-var deletion
   | TyWildcard (Just (TyVar v)) <- ty, v == var = pure (ReflCo ty)
-  -- /\ Var-wildcard deletion (same as above, but with an indirection)
-  | TyPi (Invisible _ _ r) _ <- ty, r /= Req = confesses (Impredicative var ty)
   -- /\ Predicativity checking
   | otherwise = do
+      -- guarded <- view guarded
```
SquidDev: Can we remove this (and the lines just below)?

tests/types/gadt/dict02-fail.out
```diff
@@ -1,6 +1,12 @@
...
+dict02-fail.ml[17:17 ..17:22]: error
    │ 
 17 │ let x = with_d (bar ()) () (fun x -> show x)
-   │                         ^^
-  Couldn't match actual type unit
-    with the type expected by the context, int
+   │                 ^^^^^^
+  Couldn't match actual type dict (show int)
+    with the type expected by the context, dict ('c unit)
+dict02-fail.ml[17:17 ..17:22]: error
```
SquidDev: Hrmr, this message appears to be repeated twice. Don't know if that's to be expected or not.

tests/types/rankn/impredicative01-fail.out
```diff
@@ -1,6 +1,3 @@
-impredicative01-fail.ml[6:38 ..6:52]: error
-  │ 
-6 │ let foo : id (forall 'a. 'a -> 'a) = Id (fun x -> x)
-  │                                      ^^^^^^^^^^^^^^^
-  Couldn't match actual type id ('ah -> 'ah)
-    with the type expected by the context, id (forall 'a. 'a -> 'a)
+id : type -> type
```
SquidDev: Can we re{name,do,move} this test, if it no longer fails?

src/Types/Infer/App.hs
```diff
@@ -0,0 +1,224 @@
...
+      let sp = spine fn
+       in (TypeArg arg, BecauseOf ex):sp
+    spine ex = [(ExprArg ex, BecauseOf ex)]
+
+    checkArguments ((ExprArg arg, _):as) (Quant tau dom cod inst_cont qs) =
+      case dom of
+        Anon dom -> do
+          x <- check arg dom
+
+          let cont ex = App ex x (annotation ex, cod)
```
SquidDev: `annotation ex <> annotation x` I think (and similarly in the `TyApp` case. Currently this respans applications to contain the function name, which makes a couple of errors messages more confusing.

src/Types/Infer/App.hs
zardyh: done

src/Types/Infer/App.hs
zardyh: No: this lets us individually control which modules to trace. Kinda sad, though

src/Types/Unify.hs
zardyh: Yeah

tests/types/gadt/dict02-fail.out
zardyh: One is probably from the quick look. I'm not too bothered

tests/types/rankn/impredicative01-fail.out
zardyh: sure

src/Types/Infer/App.hs
zardyh: done, too

PR #207, opened 2019-10-25
Feature: 'include' statement

zardyh:
`include` is exactly as `open` except the signature of the included module is concatenated with the signature of the including module. Concretely:

```ocaml
module X : sig val x : int end = begin
  let x = 1
end

module Y = (begin
  let y = 1
  include X
end) : sig
  val x : int
  val y : int
end
```

### DIFF
--- a/src/Parser/Token.hs
+++ b/src/Parser/Token.hs
@@ -55,7 +55,7 @@ data TokenClass
   | TcWhen -- ^ A @when@ token.
   | TcPrivate -- ^ A @private@ token.
   | TcImport -- ^ An @import@ token.
-  | TcInclude -- ^ An @import@ token.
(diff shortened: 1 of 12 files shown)


src/Parser/Token.hs
```diff
@@ -55,6 +55,7 @@ data TokenClass
   | TcWhen -- ^ A @when@ token.
   | TcPrivate -- ^ A @private@ token.
   | TcImport -- ^ An @import@ token.
+  | TcInclude -- ^ An @import@ token.
```
SquidDev:
```suggestion
  | TcInclude -- ^ An @include@ token.
```

zardyh: of course

PR #208, opened 2019-10-27
Add `typeable`

zardyh:
wowie, magic classes

### DIFF
--- a/src/Backend/Lua/Builtin.hs
+++ b/src/Backend/Lua/Builtin.hs
@@ -215,6 +215,7 @@ builtins =
           return { x, __tag = "ShowType" }
         end
       |] )
+
   , ( tcTYPEABLE, "_Typeable", [], Just (1, \[x] -> (mempty, [[lua| %x() |]]))
     , [luaStmts|
         local function _Typeable(x)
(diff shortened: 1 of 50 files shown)


src/Syntax/Implicits.hs
```diff
@@ -284,6 +286,13 @@ matches (TyPromotedCon t) (TyPromotedCon t') = t == t'
 matches TyPromotedCon{} _ = False
 
 matches (TyApp f x) (TyApp f' x') = f `matches` f' && x `matches` x'
+
+matches (TyApp f x) (a :-> b) =
+  matches f (TyApp (TyCon (TgName (T.pack "->") (negate 7))) a) && matches x b
```
SquidDev: This seems incredibly fragile. How are we doing this elsewhere?

src/Backend/Lua/Builtin.hs
```diff
@@ -215,5 +215,60 @@ builtins =
...
+                        / 10^(math.floor(math.log(math.abs(ta.fingerprint)) + math.log(math.abs(tb.fingerprint))))
+          else
+            finger = ta.fingerprint + 10^(math.floor(math.log(-ta.fingerprint))) * tb.fingerprint
+          end
+
+          return { { fingerprint = finger, name = ta.name .. " :$ " .. tb.name } , __tag = "TypeRep" }
+        end
+      |] )
+
+-- Type _Typeable_app instance makes a decimal of the concatenation of
```
SquidDev: Hrmr, can you explain? Is this a backend issue, is there anything I can do to fix it?

src/Syntax/Implicits.hs
zardyh: Exactly like this (but using the builtin name)

zardyh: (done)

src/Control/Monad/Infer.hs
```diff
@@ -485,6 +488,12 @@ instance Pretty TypeError where
   pretty (NotAClass name) =
     vsep [ "Can not make an instance of" <+> pretty name <+> "because it is not a class" ]
 
+  pretty (DIMalformedHead _) =
+    "Malformed head in" <+> keyword "deriving instance" <+> "clause"
+
+  pretty (DICan'tDerive n _) =
+    "Can't derive an instance of class" <+> stypeSkol (pretty n)
```
SquidDev: Should be "I can't derive, whether you should live or die".

PR #213, opened 2019-10-31
Allow generating Lua modules as well as programs

### LINKED ISSUE #195: Amulet and Lua interoperability
Currently, there is no easy way of using Amulet libraries from Lua. Proposals can get a little bit ugly, so I guess here some ideas can be discussed?
- Polluting the global evnironment
- `require "amuletmodule"` and a compiler flag to generate an export table

I'm also guessing there would be issues with how to handle types? It could get a bit complex, so I'm just leaving this here.

SquidDev:
Use --export to return any variables which are exported from the current program. This will also suppress DCE on these terms.

Closes #195

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ install:
 
 script:
 - stack --no-terminal build --fast --test --no-run-tests --ghc-options "-Werror -fmax-pmcheck-iterations=5000000"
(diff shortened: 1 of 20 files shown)


zardyh:
Unused constructors are still being removed from the Core program even when they're exported:
```ocaml
type t = T
```

```lua
do
  return { T = T }
end
```

SquidDev:
Good catch, thanks!

In answer to a couple of comments on Discord:
 - With regards to classes/instances, I really wasn't sure about exposing them (they have magic names, actual representation is an implementation detail etc...) so elected not to for now.

 - Yep, we're exporting curried functions for now. It's definitely not ideal, but I'd like to get some other plumbing in within Lower before changing that.

PR #220, opened 2019-11-12
WIP: Build with Jenkins

zardyh:
[here](https://callisto.amulet.works)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,11 +2,18 @@ pipeline {
   agent any
 
   stages {
+    stage('Set up stack') {
+      steps {
+        sh 'stack exec --package=hlint -- true'
+      }
+    }
+
     stage('Validate the pushed code') {
       steps {
         sh 'stack exec --package=hlint -- hlint --git'
       }
     }
+
     stage('Run tests') {
       steps {
         sh '''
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+  stages {
+    stage('Validate the pushed code') {
+      steps {
+        sh 'stack exec --package=hlint -- hlint --git'
+      }
+    }
+    stage('Run tests') {
+      steps {
+        timestamps () {
+          sh 'stack test --fast --test-arguments "--xml junit.xml --display t"'
```
SquidDev:
```suggestion
-          sh 'stack test --fast --test-arguments "--xml junit.xml --display t"'
+          sh 'env AMC_LIBRARY_PATH=$PWD/lib/ stack --no-terminal test --fast --ghc-options "-Werror -fmax-pmcheck-iterations=5000000" --test-arguments "--xml junit.xml --display t --hedgehog-tests 5000 --timeout 80'
```

Maybe? I'd also be inclined to split this into install deps, build, and then test, like we do for Travis.

zardyh: the deps are shared though

PR #225, opened 2019-11-23
Compile to native binaries via Chicken

zardyh:
TODO:

- [x] Backend-conditional imports
- [x] Generate the built-ins we need
- [x] Records via Chicken
- [x] Values


zardyh:
The Scheme it generates is utterly unreadable, but hey, it's IR. Nobody reads IR, right?

PR #243, opened 2019-12-09
Add a show instance for string.

davidgarland:
Sorry about the issue submitted and quickly closed a little earlier-- I was thinking under the assumption that a show instance for string already existed.

### DIFF
--- a/lib/amulet/base.ml
+++ b/lib/amulet/base.ml
@@ -217,6 +217,10 @@ instance show 'a => show (ref 'a) begin
   let show x = "ref (" ^ show (!x) ^ ")"
 end
 
+instance show string begin
+  let show x = "\"" ^ x ^ "\""
+end
+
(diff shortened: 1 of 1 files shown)


zardyh:
This instance won't properly escape control characters. Moreover, why do you want to convert a string to a string?

davidgarland:
I want to be able to print structures that contain strings-- I'm obviously not interested in using `show` on a string directly.

As for the control characters bit-- good point, I guess it needs some more attention to detail than that. Perhaps I can take another stab at this when it's not about time for me to get some sleep.

davidgarland:
I'll close this for now-- there's some more that needs to be added in terms of utility functions (character type predicates like `is_digit` for instance) before writing the code to do proper escaping would be elegant. In addition this doesn't seem like a large priority, so best to not keep it open indefinitely.

PR #244, opened 2019-12-11
Abstract out Backend.Lua.Emit's graph handling

SquidDev:
This simply moves the "emit graph" code into a separate module, which should allow us to more easily implement other backends which generate syntax trees (such as a JS backend, or improving the Scheme one).

In order to define a new backend, one creates an empty data type `Foo`, and defines an instance `Emitter Foo`, which determines the type of statements and expressions within the syntax, as well as how to create such terms. This is then used to build up a partial ordering of statements/expressions, and stitch them together when emitting a function.

This simplifies the Lua backend a little (reduces it by ~300 loc), though it's still annoyingly large.

We also make Core a little less polymorphic: Types, coercions and atoms are no longer polymorphic in their variable - they always store `CoVar`. We never did anything useful with this polymorphism, so we just ended up having an awful lot of boxing and unboxing.

This means that binders are the only location where variables are polymorphic. This shouldn't be an issue, as the only time we use this polymorphism (the `Occurrence` pass), the information was only correct at binding points anyway.

### DIFF
--- a/amuletml.cabal
+++ b/amuletml.cabal
@@ -373,6 +373,7 @@ library
                      , Backend.Lua.Inline
                      , Backend.Lua.Builtin
                      , Backend.Lua.Postprocess
+                     , Backend.EmitGraph
                      , Backend.Escape
                      -- Lua
                      , Language.Lua.Quote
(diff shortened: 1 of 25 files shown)


SquidDev:
I was hoping to change the Scheme backend to use this new system. However, I've not got round to it in the last week, so it feels like it's worth merging this in the current state, and we can refine it later.

PR #245, opened 2019-12-11
Refactor: Split up Repl.hs

CrazedProgrammer:
This commit attempts to split up Repl.hs into modules with their own
responsibilities. This will most likely make it easier to add new
commands and features to the REPL.

The definition of the types or functions have not been altered, they have only been moved.

I want to add a :help command, and this would make it easier for me to implement :smile:


bin/Amc/Repl/Command.hs
```diff
@@ -0,0 +1,193 @@
...
+execCommand tid "q"    _ = finish tid
+
+execCommand _ "l"    arg = loadCommand arg
+execCommand _ "load" arg = loadCommand arg
+
+execCommand _ "r"      _ = reloadCommand
+execCommand _ "reload" _ = reloadCommand
+
+execCommand _ "t" arg = typeCommand arg
+execCommand _ "type" arg = typeCommand arg
```
zardyh: Align these (on the `=`)

bin/Amc/Repl/Command.hs
```diff
@@ -0,0 +1,193 @@
...
+
+execCommand _ "i" arg = infoCommand arg
+execCommand _ "info" arg = infoCommand arg
+
+execCommand _ "c" arg = compileCommand arg
+execCommand _ "compile" arg = compileCommand arg
```
zardyh: These too

bin/Amc/Repl/Command.hs
```diff
@@ -0,0 +1,193 @@
...
+  (_, completions) <- completeInScope (reverse word, "")
+  out <- gets outputHandle
+  liftIO $
+    for_ completions $ \(Completion rep _ _) ->
+      hPutStrLn out rep
+
+execCommand _ cmd _ = outputDoc ("Unknown command" <+> verbatim cmd)
+
+loadCommand :: (MonadState ReplState m, MonadIO m) => String -> m ()
+loadCommand arg = case parseArgs arg of
```
zardyh: Line break after the `=` here

PR #247, opened 2019-12-19
Add support for dot-op assignments

SquidDev:
This now allows one to write expressions like:

```ml
let x = Array.create (fun _ _ -> "") (10, 10)
x.(0, 0) <- "Something fancy"
```

(ignoring the fact that we have no array library yet).


zardyh:
We do have an array library and it doesn't look like that. Thanks :+1:

PR #252, opened 2019-12-25
Add :help REPL command with explanations of every command

CrazedProgrammer:
This PR adds a `:help` / `:?` command that displays what each command does.
When starting the REPL, this fact is also displayed.

This would help new users get a hang of the REPL and the available commands.

### DIFF
--- a/bin/Amc/Repl/Command.hs
+++ b/bin/Amc/Repl/Command.hs
@@ -60,6 +60,9 @@ execCommand :: (MonadState ReplState m, MonadIO m) => Listener -> String -> Stri
 execCommand tid "quit"  _   = finish tid
 execCommand tid "q"     _   = finish tid
 
+execCommand _ "h"       _   = helpCommand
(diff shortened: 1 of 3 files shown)


bin/Amc/Repl/Command.hs
```diff
@@ -60,6 +60,9 @@ execCommand :: (MonadState ReplState m, MonadIO m) => Listener -> String -> Stri
 execCommand tid "quit"  _   = finish tid
 execCommand tid "q"     _   = finish tid
 
+execCommand _ "?"       _   = helpCommand
+execCommand _ "help"    _   = helpCommand
```
SquidDev: Maybe also just `h`?

CrazedProgrammer: Sure. I think `:?` looks better, but `:h` is more consistent.

SquidDev:
And merged! Thank you very much. ❤

PR #253, opened 2019-12-28
Add support for -Werror and friends

### LINKED ISSUE #228: Add flag to treat warnings as errors
Would be nice to be able to error out if there are any warnings.

SquidDev:
We currently support the following flags:
 - `-Werror` (error on all warnings/notes)
 - `-Wno-error` (disable erroring on warnings)
 - `-Werror=3002` (always error on 3002)
 - `-Wno-error=3002` (don't error on 3002, if 3002 is normally a warning/note).

Note that these flags are applied in order, so `-Werror -Wno-error` will do nothing.

Ideally we'd also support error names, as this doesn't allow you to error on messages without codes, but it's a good start.

Closes #228.


SquidDev:
Sorry for doing all the boring stuff, but it was something I knew I could get done quickly.

SquidDev:
Of course.

PR #261, opened 2020-01-13
add a library of stream functions

plt-abigail:
```ocaml
let primes =
  let open Seq in
  let rec sieve p =
    let (p, xs) = from_option (lazy (error "wasn't infinite")) (uncons p)
    p :> lazy (xs |> filter ((> 0) % flip mod p) |> sieve)
  sieve (arithmetic_seq 2 3)

let factor i =
  let rec go acc primes i =
    match i, Seq.uncons primes with
    | 1, _  -> acc
    | p, None -> p :: acc
    | i, Some (p, _) when i `mod` p == 0 ->
        go (p :: acc) primes (i // p)
    | i, Some (_, ps) -> go acc ps i

  go [] (primes |> Seq.take_while (fun p -> p <= int_of_float (i / 2))) i
```

### DIFF
--- a/lib/data/stream.ml
+++ b/lib/data/stream.ml
@@ -1,0 +1,265 @@
+open import "prelude.ml"
+
+(* The step type *)
+private type step 's 'a =
+  | Yield of 's * 'a
+  | Skip of 's
+  | Done
+
+instance eq 's * eq 'a => eq (step 's 'a) begin
+  let a == b =
+    match a, b with
+    | Yield (s, x), Yield (s', x') -> x == x' && s == s'
+    | Skip s, Skip s' -> s == s'
+    | Done, Done -> true
+    | _, _ -> false
+end
+
+(** The abstract type of streams. *)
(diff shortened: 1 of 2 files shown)


lib/data/stream.ml
```diff
@@ -0,0 +1,265 @@
+open import "prelude.ml"
+
+(* The step type *)
+private type step 's 'a =
```
SquidDev: Hrmr, we need a way to have abstract types which are still public.

lib/prelude.ml
```diff
@@ -29,3 +29,4 @@ let print = Io.print
 
 let flip f x y = f y x
 let x |> f = f x
+let f # g = fun x -> f (g x)
```
SquidDev: Hrrm, is this not the same as `%>` in data/category.ml?

plt-abigail: But specialised to `(->)` so we don't need `Category` dictionaries

PR #262, opened 2020-01-14
Add range syntax

plt-abigail:
`[ x .. y ]` is syntax for `range x y`
`[ x, y .. z ]` is syntax for `range_then x y z`


plt-abigail:
Update, with ranges generalised and owing to SquidDev's very nice `let open` shorthand, you can write a range `stream` like so:

```ocaml
let ints = S.[ 1, 2 .. 10 ]
```

PR #270, opened 2020-02-03
Include the library path on missing imports

SquidDev:
- Importers now return a "SearchedIn" result, which details where we attempted to load a file from. This is then printed in error messages.

 - Refactor some common library loading code into `Frontend.Driver` from the LSP server and main compiler driver.

Example output:
```
tests/amc/files/import.ml[1:6 ..1:23]: error (E1010)
  Cannot resolve "my_lib.ml"
  │ 
1 │ open import "my_lib.ml"
  │      ^^^^^^^^^^^^^^^^^^
  
  Searched in:
    • /home/squiddev/programming/amulet/lib/my_lib.ml
    • /home/squiddev/programming/amulet/.stack-work/blah/blugh/lib/my_lib.ml
    • /home/squiddev/programming/amulet/.stack-work/blah/lib/my_lib.ml
```

### DIFF
--- a/amuletml.cabal
+++ b/amuletml.cabal
@@ -286,6 +286,7 @@ library
                      -- Main frontend
                      , Frontend.Driver
                      , Frontend.Errors
+                     , Frontend.Files
                      -- the AST
                      , Syntax
                      , Syntax.Let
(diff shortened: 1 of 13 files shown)


SquidDev:
Mostly done so I could actually understand why my lsp configuration was broken. But I still think it's useful.

The previous error message in Emacs

SquidDev:
Tests should be working as of 26ec59db6b6c9e45974ce2e779f572bf4b282e59 - it's not a problem with this PR.

PR #272, opened 2020-02-12
Rework annotations

SquidDev:
- Split annotations in to three kinds:
   - `Ann p`: Our existing annotation. Stores positions, then types of      positions.
   - `RawAnn p`: Used for top-level statements. Always stores positions.
   - TypeAnn p: Used for types. Stores positions, and then disappears     during desugaring.

 - Names within types (tyvars, type constructors, and promoted constructors) now track their positions.

 - Split up "annotation extraction" and "position extraction" into two separate passes. This removes the need to use generics to extract an expression's type when lowering.

 - Correct the span of some nodes, to more accurately reflect the full contents.


SquidDev:
I'm aware this is a horror-story of a commit, and does make the TC a bit uglier. Wasn't really sure of an appropriate solution though.

PR #273, opened 2020-02-13
A terrible hash map implementation

SquidDev:
- Add several type classes for "indexable" types.
   - `index`: This provides `.()` and `.?()` getters.
   - `mut_index`: This provides a mutating setter `.[]<-`.
   - Add a `set_index` type class. This provides a setter `.()<-`, which returns an updated collection.
 - Add `hashable` type class (in `data/hash.ml`), and instances for the basic types.
 - Add a primitive mutable hash map (in `data/hash/map.ml`). There's a lot of work which could be done to improve this, but it's a good enough start for basic functionality.


plt-abigail:
I wonder if a HAMT implementation wouldn't be better?

SquidDev:
> But I guess it doesn't matter if everything's out of bounds for it..

Yeah, I'm _fairly_ sure this is a safe optimisation as empty array can't be modified or indexed. Java does it, though I realise that's hardly a great commendation.

> I wonder if a HAMT implementation wouldn't be better?

I don't know if it's worth having both? I definitely want some kind of mutable map structure, though am not exactly wedded to this particular design - might try out something closer to Lua's hash implementation.

davidgarland:
If performance is your aim, I might recommend looking into [robin hood hashing](https://www.sebastiansylvan.com/post/robin-hood-hashing-should-be-your-default-hash-table-implementation/) with [backwards shift deletion](http://codecapsule.com/2013/11/17/robin-hood-hashing-backward-shift-deletion/), more or less keeping the current model of using a flat array. (Though, I think you ought to make the key/value pair into two individual arrays for the usual SoA benefits-- `<$>` on hashmaps would then no longer need to load the keys into cache as well, for instance, and the same goes for iterating over keys only.)

SquidDev:
Thanks for the links David! I was vaguely aware of robin hood hashing, but those are articles were very useful.

I must confess this implementation was largely written from whatever I remember of algorithms courses, so is definitely not as sane as it should be. I definitely want to replace this with a better implementation.

That said, performance is a bit tricker - the backend spits out absolutely terrible Lua code, which means any memory optimisations are overshadowed by the number of closures/tables we're allocating for every call. I really need to fix this before doing any serious profiling:

```lua
-- map.[1] compiles to:
local tmp0 = _dot_28_290({
  hash = function(x) return x end,
  ["Hashable$agcy"] = {
    ["=="] = int_eq0,
    ["<>"] = function(bqq)
      return function(bqr)
        return bqq ~= bqr
      end
    end
  }
})(map)(1)
```

Looks like it's time to do some optimiser work again :D:.

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
+open import "../prelude.ml"
+
+class eq 'a => hashable 'a
```
plt-hokusai, line 3: `begin`/`end` please

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
...
+  val hash : 'a -> int
+
+instance hashable int
```
plt-hokusai, line 6: ditto

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
...
+  let hash x = x
+
+instance hashable ()
```
plt-hokusai, line 9: ditto

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
...
+  let hash () = 0
+
+instance hashable bool
```
plt-hokusai, line 12: ditto

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
...
+  let hash x = if x then 1 else 0
+
+instance hashable 'a * hashable 'b => hashable ('a * 'b)
```
plt-hokusai, line 15: ditto

lib/data/hash.ml
```diff
@@ -0,0 +1,25 @@
...
+  let hash (a, b) = 31 * hash a + hash b
+
+instance hashable string
+  let hash xs =
+    (* Hash using the djb2 hash function *)
+    let open import "../lua/string.ml"
+    let rec go x = function
+    | 0 -> x
```
plt-hokusai, line 23: Shouldn't we take the remainder of this by a suitable prime?

plt-hokusai: Moreover, this algorithm works to hash any Foldable, does it not? We should at least add instances for lists.

lib/data/array.ml
```diff
@@ -111,14 +118,43 @@ instance eq 'a => eq (array 'a) begin
...
 
-let ( .() ) (Array r as arr) i =
-  in_bounds "(.())" arr i
-  geti r.backing (i + r.offset)
-
-let ( .()<- ) (Array r as arr) i x =
-  in_bounds "update" arr i
-  let _ = seti r.backing (i + r.offset) x
-  ()
+let empty = Array { length = 0, offset = 0, backing = empty_storage }
```
plt-hokusai, line 121: I really don't feel safe sharing a single table as backing storage for a module of mutable arrays.

lib/data/hash/map.ml
```diff
@@ -0,0 +1,84 @@
...
+      | Nil -> None
+      | Cons { k = k', v } when k == k' -> Some v
+      | Cons { next } -> find next
+    let h = (hash k .&. (Array.size array - 1)) + 1
+    find (array.[h])
+
+let iter f (HashMap container) =
+  let Container { array } = !container
+  let rec iter_one = function
+  | Nil -> ()
```
plt-hokusai, line 47: Could you indent these arms?

lib/data/hash/map.ml
```diff
@@ -0,0 +1,84 @@
...
+let load_factor = 2
+
+let make () = HashMap (ref (Container { array = Array.empty, size = 0, initial = 16 }))
+
+let make_with_capacity c =
+  let initial = if c <= 1 then 1 else
+    Math.log (float_of_int c) /. Math.log 2.0 |> Math.ceil |> (2**)
+  HashMap (ref (Container { array = Array.empty, size = 0, initial }))
+
+instance hashable 'k => index (t 'k 'v)
```
plt-hokusai, line 30: Begin/end here too!

lib/data/hash/map.ml
```diff
@@ -0,0 +1,84 @@
...
+  | Nil
+  | Cons of { k : 'k, v : 'v, next : pairs 'k 'v }
+
+private type container 'k 'v = Container of {
+  array : Array.array (pairs 'k 'v),
+  size : int,
+  initial : int,
+}
+
+type t 'k 'v = private HashMap of ref (container 'k 'v)
```
plt-hokusai, line 19: This type is missing instances for `functor`, `foldable`, and `traversable`.

lib/data/map.ml
```diff
@@ -87,17 +88,24 @@ let lookup k (M tree) =
       | Lt -> go l
       | Gt -> go r
   go tree
-let (.?[]) map key = lookup key map
 
-(** Lookup a key on the map. If it is not present, raise an exception.
- *)
-let (.[]) map key =
-  match lookup key map with
+(* instance ord index map where *)
```
plt-hokusai, line 92: Huh?

lib/prelude.ml
```diff
@@ -14,7 +14,7 @@ include import "./data/enumeration.ml"
 
 module Io = import { lua = "./lua/io.ml" }
 module Bit = import { lua = "./lua/bit.ml" }
-open Bit.Ops
+include Bit.Ops
 
 open import { lua = "./lua/conversion.ml" }
```
plt-hokusai, line 19: Include this one too, if you don't mind.

src/Backend/Lua/Emit.hs
```diff
@@ -835,3 +835,5 @@ opOfIntrinsic StrConcat = ".."
 opOfIntrinsic StrEq = "=="
 opOfIntrinsic StrLt = "<"
 opOfIntrinsic StrLe = "<="
+
+opOfIntrinsic BoolEq = "=="
```
plt-hokusai, line 839: I wonder what code we would generate for
```ocaml
let a == b =
  match a, b with
  | true, true | false, false -> true
  | _ -> true
```
(no need to act on this)

lib/amulet/base.ml
```diff
@@ -29,6 +29,7 @@ end
 instance eq int    begin let (==) = Base_defs.int_eq end
 instance eq float  begin let (==) = Base_defs.float_eq end
 instance eq string begin let (==) = Base_defs.string_eq end
+instance eq bool   begin let (==) = Base_defs.bool_eq end
```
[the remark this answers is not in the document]
plt-hokusai: Nvm - I saw the primitive way down there. But `ord bool` is still good to have!

lib/data/array.ml
SquidDev: I get that, but it is perfectly safe - you can't mutate it as it's empty.

lib/data/array.ml
```diff
@@ -111,14 +118,43 @@ instance eq 'a => eq (array 'a) begin
...
+
+let iter f (Array { length, offset, backing }) =
```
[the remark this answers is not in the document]
SquidDev: Hmm. Any reason why? OCaml generally has the object you act on as the last argument, as it's more amenable to currying/`|>`.

lib/data/hash/map.ml
SquidDev: Good point. I originally left it off, as functors on mutable types feels a little  weird, but we also do in for arrays after all.

lib/amulet/base.ml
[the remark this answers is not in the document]
SquidDev: Yeah, the generated code will be bad (similarly for any enum type I guess). Hence we cheating with an intrinsic for now.

lib/data/array.ml
```diff
@@ -10,11 +11,14 @@ type array 'a =
     backing : storage 'a
   }
 
-let private in_bounds str (Array r) i =
-  if (r.offset + i <= r.length) && (i >= 1) then
-    ()
-  else
-    throw (Invalid (str ^ ": index " ^ show i ^ " is out of bounds"))
+let private is_in_bounds i (Array { offset, length }) =
+  (* Flip the argument order so the pattern match happens last. Otherwise this
```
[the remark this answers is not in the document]
SquidDev: The desugaring is correct, it's just the optimiser is dumb. ` is_in_bounds (Array { offset, length }) i` desugars to:

```ml
let is_in_bounds = λ a. 
  let a' = coerce a
  match a' with
  | { offset, length } -> λ i. (* ... *)
```

And the uncurrying optimiser isn't smart enough to deal with this.

lib/data/array.ml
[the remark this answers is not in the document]
SquidDev: Also, given `iter f x` is just `foldl (fun () -> f) () x`, it feels more consistent to keep it before.

lib/data/array.ml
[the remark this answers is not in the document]
plt-hokusai: Right, I feel we should float all the lambdas outwards so we don't end up with this "problem", especially for functions defined in a single block like that

lib/data/array.ml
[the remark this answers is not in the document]
plt-hokusai: So we can write
```
iter x @@ fun e ->
   ...
```

lib/data/array.ml
plt-hokusai: Aye, if you say so!

lib/data/array.ml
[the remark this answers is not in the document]
SquidDev: `x |> iter @@ fun e ->` should work correctly.

lib/data/hash/map.ml
plt-hokusai: foldable` at least makes a lot of sense, but I think a copying `map` is worth having, at least.

lib/data/array.ml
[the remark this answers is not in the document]
plt-hokusai: Well, if the parser agrees with this now—Last I tried, I ran into a bunch of context weirdness. Admittedly it was quite a while ago.

lib/data/hash/map.ml
```diff
@@ -0,0 +1,133 @@
...
+private module Math = import "../../lua/math.ml"
+private module Array = import "../array.ml"
+
+private type pairs 'k 'v =
+  | Nil
+  | Cons of { k : 'k, v : 'v, next : pairs 'k 'v }
+
+instance functor (pairs 'k) begin
+  let f <$> x =
+    match x with
```
plt-hokusai, line 14: Sorry to nitpick like this, but do you think you could rewrite this with a local worker? To make the dictionary non-recursive

lib/data/index.ml
```diff
@@ -0,0 +1,43 @@
+open import "../prelude.ml"
+private module String = import "../lua/string.ml"
+
+class index 'a
```
plt-hokusai, line 4: missing begin/end

lib/data/index.ml
```diff
@@ -0,0 +1,43 @@
...
+  val ( .() ) : 'a -> key 'a -> value 'a
+  let ( .() ) xs k =
+    match xs.?(k) with
+    | Some x -> x
+    | None -> Exception.(throw (Invalid ".()"))
+
+  (** Look up an index, returning [None] if it is unavailable. *)
+  val ( .?() ) : 'a -> key 'a -> option (value 'a)
+
+instance index string
```
plt-hokusai, line 18: ditto

lib/data/index.ml
```diff
@@ -0,0 +1,43 @@
...
+    String.char_code_at str i |> Some
+
+  let ( .() ) str i =
+    if i <= 1 || i > String.length str then
+      let open Exception in
+      throw (Invalid ("string.(): index " ^ show i ^ " is out of bounds"))
+    else
+      String.char_code_at str i
+
+class index 'a => set_index 'a
```
plt-hokusai, line 33: ditto

lib/data/index.ml
```diff
@@ -0,0 +1,43 @@
...
+  (** Create a copy of this object with the index added or modified. *)
+  val ( .()<- ) : 'a -> key 'a -> value 'a -> 'a
+
+class index 'a => mut_index 'a
```
plt-hokusai, line 37: ditto

PR #277, opened 2020-02-16
Add :explain command to the repl

### LINKED ISSUE #276: amc explain from the toplevel
So, Amulet's errors are pretty informative, and the suggestion to use `amc explain` to get additional detail is very nice.  That said, it's a bit weird to be in the toplevel and be given the suggestion to use a separate command line tool to get at the information.  

When getting an error in the toplevel saying `try 'amc explain 1001' to see it` my first instinct was to see if `amc` was a valid function so that `amc explain 1001`  would be valid in the toplevel as well as the command-line, immediately followed by checking `:help` to see if it had any sort of `:explain` command to get at the information.  When neither worked, I went looking for something like `open import "amc.top"` or `open import "repl.ml"` to see if Amulet has some sort of toplevel helpers module that isn't loaded by default, but saw none.

Having some way to get at this information without swapping to a different terminal would be nice.

SquidDev:
This has been irking me for a while, and someone filed an issue so I added it. Fixes #276.

### DIFF
--- a/bin/Amc.hs
+++ b/bin/Amc.hs
@@ -22,7 +22,7 @@ import Text.Pretty.Semantic hiding (empty)
 import Text.Read
 
 import qualified Frontend.Driver as D
-import Frontend.Errors
+import Frontend.Errors hiding (Repl)
 
 import qualified CompileTarget as CT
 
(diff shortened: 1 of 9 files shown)


plt-abigail:
Seems simple and good

Issue #205, opened 2019-10-23
Special-casing externals


SquidDev:
As of #184, most functions were moved from the builtin list to the prelude. While this is undoubtedly a good thing, it does mean the optimiser is no longer aware of how these functions behave, and so arity information is lost, and constant folding does not occur.

@zardyh has suggested just performing constant folding in the backend, and that may be worth doing as well. However, really the optimiser needs to have some info in order to generate sane-ish code.

I think the cleanest solution to this is twofold:

 - Go the OCaml route, and define several "sorta-builtins", such as `%int_add`. We can then special-case these strings in specific places: verify can ensure declared types match up with the expected one, and optimiser can constant-fold them, and the backend can just convert them to operands.

   It's [not _especially_ elegant](https://github.com/ocaml/ocaml/blob/72ea849d2a16de0abb42afd85c014cb136822e1f/lambda/translprim.ml#L111), but I think it's our most effective option.

 - Allow annotating external functions as `[@@pure]` - this would only be used by the arity analyser to say that this function can be executed without side effects. Verify could also technically use this to warn when discarding the results of pure functions, but that's less important.

PR #279, opened 2020-02-20
Constant folding, again

SquidDev:
- Add support for "intrinsics" (see #205) to Core. These are just written as foreign bindings, prefixed with a `%` (e.g. `external val ( + ) : int -> int -> int = "%int.add"`).
 - Verify checks intrinsics exist and have the correct type.
 - Adds back the constant folding pass removed in #184.

### DIFF
--- a/src/Core/Optimise/Reduce/Fold.hs
+++ b/src/Core/Optimise/Reduce/Fold.hs
@@ -77,7 +77,9 @@ float :: Double -> Maybe (Term a)
 float = atom . Lit . Float
 
 bool :: Bool -> Maybe (Term a)
-bool x = atom (Lit (if x then LitTrue else LitFalse))
+bool x = 
+  let val = if x then LitTrue else LitFalse
+  in val `seq` atom (Lit val)
 
 atom :: Atom -> Maybe (Term a)
 atom = Just . Atom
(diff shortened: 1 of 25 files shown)


src/Core/Optimise/Reduce/Fold.hs
```diff
@@ -0,0 +1,83 @@
...
+int = atom . Lit . Int
+
+str :: T.Text -> Maybe (Term a)
+str = atom . Lit . Str
+
+float :: Double -> Maybe (Term a)
+float = atom . Lit . Float
+
+bool :: Bool -> Maybe (Term a)
+bool x = atom (Lit (if x then LitTrue else LitFalse))
```
plt-abigail: can you put that `if` in a `where`  bound variable and `seq` it?

SquidDev: Hrmr, I'm not sure where the `seq` would go? Something like this:

```haskell
bool x = val `seq` atom (Lit val) where
  val = if x then LitTrue else LitFalse
````

plt-abigail: Yeah, like that. It's probably not necessary but will probably be better for memory usage. Just stuck out to me, sorry.

PR #296, opened 2021-04-02
Mitigate exponential behaviour of inhabitance checker

### LINKED ISSUE #294: Compiler eating memory infinitely
I've tried to minimize this and I just can't seem to get the same behavior when I do.

The project is at https://gitdab.com/s/ootAI. I suspect that the `equipment` record in `oot/equipment.ml` is the problem.

If `equipment: u16` is replaced with `equipment: equipment` [here (permalink)](https://gitdab.com/s/ootAI/src/commit/1628d2ea5598cad0567606ef0aa135b822022c78/oot/item_equips.ml#L16), the compiler gets stuck and eats up all my memory.

SquidDev:
This adds a simpler version of the inhabitance checker in pmcheck. Unlike pmcheck's, this does not use any contextual information, which means it can be memoized.

There are some cases where the exponential behaviour is still present. For instance, consider the chain of sum types:

    type t1 'a = T1 of 'a
    type t2 'a 'b = T1L of (t1 'a) | T1R of (t1 'a)
    type t3 'a 'b = T3L of (t2 'a) | T3R of (t2 'a)
    (* ... *)

We cannot statically tell if tn is inhabited for any type variable. I guess we could compile this to a predicate based on the type variables, but that may be going a bit far :).

Fixes #294

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,8 +1,8 @@
 import qualified System.Environment as E
 :seti -XOverloadedStrings -XFlexibleContexts
 
-:def r! const (pure (unlines [ ":!tools/repl.sh reload", "::r", ":!tput bel" ]))
-:def reload! const (pure (unlines [ ":!tools/repl.sh reload", "::r", ":!tput bel" ]))
+:def! r! const (pure (unlines [ ":!tools/repl.sh reload", "::r", ":!tput bel" ]))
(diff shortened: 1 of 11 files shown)


plt-amy (approved):
looks good to me

